PR #17, opened 2016-01-22
Make ipcluster start and stop more robust

neurodroid:
We had some issues with ipcluster start and stop on our server. This patch makes the process a bit more robust.


agiovann:
dear Christoph,

thank you so much for your patch!

Andrea

PR #21, opened 2016-02-15
Smallfixes

j-friedrich:
Hi Andrea,
great progress on the python implementation! Finally I could gladly switch away from matlab.

Using your code I stumbled upon and fixed few issues:
- nnls, sometimes called in lars_regression_noise, needs some vector as arg, hence the ravel in line 610 of spatial.py. Also, nnls returns a tuple, of which only the first element is expected. (consider using HALS instead of scipy's nnls, which is faster in my experience, if speed is an issue)
- In nb_view_patches the numbering of shapes and activities didn't match. Fixed in lines 864, 870 in utilities.py
- I added an argument to plot_contours (line 469, 525 in utilities.py) to set the color, cause the default blue can sometimes hardly be seen. 

My editor did some PEP8 auto-formatting and inserted/deleted some spaces, hence apparently a lot of lines changed, though I have pointed out the ones going beyond proper formatting above.
Cheers
-- j

### DIFF
--- a/ca_source_extraction/spatial.py
+++ b/ca_source_extraction/spatial.py
@@ -6,7 +6,7 @@ Created on Wed Aug 05 20:38:27 2015
 """
 import numpy as np
 #from scipy.sparse import coo_matrix as coom
-from scipy.sparse import coo_matrix ,csc_matrix      
+from scipy.sparse import coo_matrix, csc_matrix
 from scipy.sparse import spdiags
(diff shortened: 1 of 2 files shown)


agiovann:
Thanks a lot for the contribution Friedrich! We implemented our own hals but it is not as fast as NMF. Would you suggest a particularly efficient  HALS implementation?

Thank you

PR #23, opened 2016-02-29
Faster hals

j-friedrich:
Hi Andrea,
this is a (~5x) faster version of HALS that alternates between updating shapes of all neurons and activities of all neurons, instead of looping over neurons in the outer loop. It avoids computation of the residual, thus being more memory efficient.

There has also been a bug in update_temporal_components, which returned the same g for groups of neurons.

Best,
Johannes


epnev:
Thanks Johannes!

PR #28, opened 2016-06-24
Allow ipcluster call from Windows

neurodroid:
This is required so that ipcluster can be called from Windows machines. Requires 4 backslashes as path separators when called from Windows, e.g. `ipcluster="C:\\\\Anaconda2\\\\Scripts\\\\ipcluster.exe"`

### DIFF
--- a/ca_source_extraction/utilities.py
+++ b/ca_source_extraction/utilities.py
@@ -12,6 +12,7 @@ from matplotlib import pyplot as plt
 from pylab import pause
 import sys
 import os
+import shlex
 try:
     import bokeh
     import bokeh.plotting as bpl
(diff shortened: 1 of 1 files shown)


agiovann:
Hi Christoph, thank you for the contribution. Since we were in the middle of a big merge and the changes did not work out of the box  I introduced your changes manually. Also, since your changes were not applying directly on linux. I patched it with a very unorthodox if condition. Would you care to check that it works for you? The problem for linux is that it would not accept the command as 
['ipcluster', 'start', '-n', '8']
but for some reason only as 
['ipcluster start -n 8']
God knows why, and since I cannot test on windows I am not sure this would be the case there too. Let me know

PR #34, opened 2016-09-27
Fixing some installation issues.

valentina-s:
Hey @agiovann, 

this PR fixes some installation problems:
* it moves the installation from files after the git clone (otherwise they are not available)
* it forces installation of pyqt4 (seems the new version of conda installs qt5 automatically)
* it fixes some error with n_processes due to change in the parameters in options

Still the build won't pass because there is an error with the tests. But I am getting this error locally too, so it is not travis. It is something about IPPP profiles. Could you try to see if you have the same error locally with nosetests?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,4 @@ branches:
     - anaconda
     - docker
     - master
+    - agiovann-master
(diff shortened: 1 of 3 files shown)


agiovann:
Hello @valentina-s ,

yes, I think there is an error in the flag of the backend. if backend='local' it should not look into that. What is the file that it is running? Is it checkup/py? I am trying to understand how travis is choosing which files to run. Let me know!

valentina-s:
@agiovann The way nosetests decides which files to run it searches through your module directory and finds the ones which start like test_ ... (it also catches tests from the unittest module, and other situations - more details here http://nose.readthedocs.io/en/latest/finding_tests.html)

PR #48, opened 2016-12-30
OASIS improvements

j-friedrich:
Implements some improvements for OASIS:
 - Grow set of pools dynamically to speed up processing of long time series
 - adaptive window size for nnls to match Ca decay time
 - re-estimation of rise time for AR(2) if AR-parameters are optimized

Adds unit-tests for deconvolution methods.

Adds 'unit-tests' for (almost) entire pipeline on small simulated toydata for 2D as well as 3D. (not testing mmaping and cluster mode)

### DIFF
--- a/caiman/source_extraction/cnmf/deconvolution.py
+++ b/caiman/source_extraction/cnmf/deconvolution.py
@@ -6,9 +6,6 @@ Created on Tue Sep  1 16:11:25 2015
 from __future__ import division
 from __future__ import print_function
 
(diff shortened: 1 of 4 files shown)


j-friedrich:
Had to fight a long battle with travis/docker, but now the unittests run finally through again. It was necessary to update test_demo.py, and I replaced the entire pipeline by merely a call of the fit method.

agiovann:
Need to apply same logic to all the demos

PR #50, opened 2017-01-06
fix HALS for initialization

j-friedrich:
fixes some bug in HALS and adjusts its variable names to better align with the usual convention used in the codebase.

Also changed README.md to show the travis banner from simons not agiovann to account for the move of the repo.


agiovann:
thank you johannes. Is speed different? Was there any optimization there?

j-friedrich:
Speed is the same, but the `S[mcell, ind_pixels][S[mcell, ind_pixels] < 0] = 0` in the former code erroneously kept negative entries S. The new code also properly deals with more than 1 background component.

PR #53, opened 2017-01-27
Save temp mmemap with four string digits

Danieldemalmazet:
I've noted a small bug in the function save_memmap_each (line 109 of file mmaping.py). I would suggest instead of: base_name+str(idx), to write instead: base_name+'{:04d}'.format(idx) . The reason is that after this function is called in the main code, you do: name_new.sort(). The problem is if you have the data contained in more than 10 tiff files, then you'll have a mismatch because the order will be the following: _1,_10,_11,.._19,_2,_20,_21,...,_3.. and so on.


agiovann:
Thank you Daniel. Let me also know if you solved the memory issue please!

PR #59, opened 2017-03-03
Deleting empty background components correctly

ecobost:
Hi Andrea,

I changed the empty component deletion part in update_spatial_component() (spatial.py) to be able to delete empty background components (it couldn't do it before), but had to change the signature of the function to also return f (before only A,C and b) because I need to know how f was changed or which component was deleted. I also changed any call to this function to receive this f and assign it to whatever f is going to be used next in the code.
Also, cnmf.fit now returns an object that has property S (inferred spike traces) and g (inferred AR parameters),
You can merge this up if you like the changes or fix it in some other way


agiovann:
Hello @ecobost, thank you for the wonderful work! Apparently you were using an old version of the code or something happened along the way. I have difficulty in merging since he pull-requests  creates conflicts with my master branch. For example lines 185-190 of cnmf.py. There I had a portion in which it was possible to initialize with a boolean mask and that part got erased for some reason. It would be really nice if we could incorporate your code without affecting the current version.  Let me know how we could  proceed to merge safely.

PR #61, opened 2017-03-05
Assert that shape argument to np.ones is an int

neurodroid:
Newer versions of numpy are getting increasingly pedantic about the types of indices, to the point where they will throw a TypeError when a float is used as an index.


agiovann:
thank you @neurodroid for the contribution

PR #63, opened 2017-03-14
Some bug fixes (deleting empty background components, dealing with non-spiking temporal components, complex casted to float warning)

ecobost:
I changed three things:
1. update_spatial_components is now able to delete empty background components (changes in spatial.py). For that it need to return both b and f  (before it only returned b) so I had to change the signature and also changed every call in the repo to receive f ( changes in some places, including the demos, just receives the f now).
2. constrained_OasisAR2 won't crash now if the component had no spikes, it did before (changes in deconvolution.py). For now, I just leave the component there but maybe they could be deleted (I wouldn't, they don't spike at all but maybe they truly show something in the scan). Also I changed the spike window from np.arange(-2,2) = [-2,-1,0,1] to np.arange(-2,3) = [-2,-1,0,1,2] which seems to make more sense.
3. No 'ComplexWarning: Casting complex values to real discards the imaginary part' in constrained_foopsi_parrallel (temporal.py). Polynomial roots are in theory complex numbers but I just discard the imaginary part. That's the behavior it had anyway (discarding the imaginary part) but now it does not show the warning. I am not so sure about this one but it does not seem to matter much (and not at all if you don't provide an initial calcium concentration)

These are mostly edge cases but I found them when running on our data so I fixed them. I ran the three demos in the master directory and they work fine (other than this warning: '/data/CaImAn/caiman/components_evaluation.py:116: RuntimeWarning: divide by zero encountered in log', but I think that was there before).

Cheers,


agiovann:
Beautiful! 

Thanks a lot!

Andrea

PR #64, opened 2017-03-17
Another set of small bug fixes

ecobost:
Added ability to create more than one background component (just sending the correct number to initialize_components).
Added ability to not receive gSig (used in greeedy_roi) when not using greedy_roi.
Thresholding each spatial mask so that by default (nrgthr=0.999) everything that is not zero is considered active and the connected component with biggest energy is loooked for there. As it was before nothing was set to active so the entire image was a single connected component. For greedy roi and neurons results in the same behavior, for sparse nmf it now selects the biggest (or highest energy) connected component among all the sparse ones (otherwise all components covered the entire image and there was a lot of overlap)
Changed most absolute imports to relative imports to run it as a module (no need to install)


agiovann:
Hello @ecobost , some tests are failing. Does it have to do with the pull request being incomplete?

ecobost:
Hi @agiovann Actually not, this version works fine for me. The errors are in a couple of tests where the expected answer is 32282000 and the obtained answer is 31971523.39683021. It does produce the same answer for both single thread and parallel (part of the test) but not the one the test was looking for (32282000). I don't  really know why is that, though, it probably has to do with how I threshold the image at 0.9999 of the energy (before it was effectively 1) maybe discarding some pixels that before made the cut (the entire image made the cut) but I am not sure about this because all other tests pass

Right now I am trying to run in a 512 x 512 scan with 35 000 timepoints and having some trouble but I think that's on our side (either the filesystem or thread management) so I wasn't planning to push anything else for now

PR #66, opened 2017-03-30
Make cnmf_patches work with newer versions of numpy (tested in numpy 10.1.4 and 12.1.0)

ecobost:
Indexing a memory mapped array with a list (such as 'Yr=Yr[idx_,:]' in map_reduce.cnmf_patches()) loads the array in memory and returns an ndarray. Now it slices the mmap array: this preserves it in disk and gives you a view of the memory mapped array (changes in map_reduce.cnmf_patches()). Older versions of numpy would work (because it returns the array as mmap, althugh it is actually loaded in memory) but consume more memory.
Cheers,


ecobost:
the travis build failed while building the docker image. @agiovann could you rerun it just to be sure. I have checked all demos in my own docker container.

agiovann:
Yes, you are definitively right. From the release documentation:

"Operations on np.memmap objects return numpy arrays in most cases
Previously operations on a memmap object would misleadingly return a memmap instance even if the result was actually not memmapped. For example, arr + 1 or arr + arr would return memmap instances, although no memory from the output array is memmaped. Version 1.12 returns ordinary numpy arrays from these operations."

So. apparently nothing should change! Thanks a lot for catching this one!

PR #72, opened 2017-05-03
Fix bugs with newer version of bokeh & Python 3's urllib

jordan-g:
This commit fixes two bugs:

1. Python 3 replaces `urllib2` with `urrllib.request` and `urllib.error`. `urllib.request.urlopen()` in Python 3 is equivalent to `urllib2.urlopen()` in Python 2.

2. Newer versions of the bokeh package don't use `vform()` & `hplot()`, instead they use `row()` and `column()`, which seem to work equivalently.

### DIFF
--- a/caiman/utils/utils.py
+++ b/caiman/utils/utils.py
@@ -9,15 +9,18 @@ from __future__ import print_function
 #%%
 import numpy as np
 import os
-import urllib2
+try:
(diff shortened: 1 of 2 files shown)


jkobject:
As I tried the plottings of the visualisation.py file : 
and it doesn't seems to work
capture d ecran 2017-06-20 16 36 23

PR #75, opened 2017-05-15
Support large sima datasets

neurodroid:
Sima sequences are read in sets of 1000 frames to avoid memory errors.


agiovann:
Also observe that now can support large h5 files as used by SIMA for both rigid and piecewise rigid motion correction

PR #79, opened 2017-06-15
Kalfon

jkobject:


jkobject:
renaming some import for movies.py
changing comments, removing a test and the ''' into """ for rois
preprocessing, only changing the location of the functions

jkobject:
oups something went wrong with the nosetests. and I forgot to add assertions.

jkobject:
this is an error that I am not able to understand.

jkobject:
I have removed only commented parts.(approved by Andrea)

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -451,51 +469,54 @@ def greedyROI(Y, nr=30, gSig=[5, 5], gSiz=[11, 11], nIter=5, kernel=None, nb=1):
...
+    @Author: Eftychios A. Pnevmatikakis based on a matlab implementation by Yuanjun Gao
             Simons Foundation, 2015
+
+    See Also:
+    -------
+    http://www.cell.com/neuron/pdf/S0896-6273(15)01084-3.pdf
+
+
     """
-    debug_ = False
```
agiovann: why did you erase this portion?

agiovann: OK, it is not super important

caiman/utils/utils.py
```diff
@@ -22,22 +32,45 @@
...
-        if not(os.path.exists('./example_movies/demoSue2x.tif')):        
-            url = 'https://www.dropbox.com/s/09z974vkeg3t5gn/Sue_2x_3000_40_-46.tif?dl=1'
-            print("downloading demo Sue2x with urllib")
-            f = urlopen(url)
-            data = f.read()
-            with open("./example_movies/demoSue2x.tif", "wb") as code:
-                code.write(data)
-        else:
-            print('File already existing')
+        for f in file_list:
```
agiovann: here I would not force to download all the files if not required...

jkobject:
Seen your comment, 
I will prepare a reply for tomorrow. 
Most of it is just entirely due to old commits that I have not managed well 
when you ignore files with gitignore but you have already pushed them before.
Will repair all of that.
Have a good night.

jkobject:
everything should be fine, I have manually removed the files that haven't due to the git ignore.
I had already taken in account johannes change, I don't see any modification in the install file except for two comments that help the user and the shell understand what it is.

jkobject:
It should work better now.

j-friedrich:
Hi Jeremie. I made some more cumbersome changes to visualization.py to use the fairly recent Bokeh array protocol. That enabled to fix the problem with the max-projected plot for 3D data and made indeed all Bokeh plots much faster to plot and interact. Committed directly to master, you might have to check whether your code is compatible and merge my changes to yours.

PR #98, opened 2017-10-06
nasty bug and Windows installation instructions

meChiel:
Hey,
 
I tried the OnAcId demo on a windows machine. Very nice.
1.I saw the installation procedure was outdated so did an update.
2.I also found an enoying error in the algortihm when I wanted to initialise more than 200 compounds. I (after some serious debugging) resloved this problem.

Keep up the good work


agiovann:
Thank you, wonderful job and great catch!

PR #107, opened 2017-10-19
Bugs squashed

HagaiHargil:
1. In the main pipeline I've added a call to the DF/F calculation function.
2. I started doing preparations for slice-based readout of files. This is important for cases in which a stack is interleaved with more than one spectral channel. The online pipeline already has support for this, I'm trying to bring this one on par. It's still not ready, though.
3. I commented out a line in `components_evaluation.py` which causes bugs (the first change below). It seems like no one is using this `significant_samples` variables, so I allowed myself to do it. It was hard to reproduce, so I can't tell exactly what caused the issue, but generally, the error was caused by the fact that sometimes `signifcant_samples__` contained a heterogeneous list of arrays and numbers, which meant that `np.concatenate` couldn't simply coerce it into an array. The problem was that this didn't happen with every stack - and I couldn't find the pattern connecting the stacks that were affected. 
4. When saving a stack with Fiji the `key` keyword in `tifffile.imread` is pretty useless, since there's only one key. Instead, if you want to read only a substack of the full file - you have to first load it into memory and then slice it. I couldn't find a better way which supports both Python versions, and it's only a problem if the stack was generated by ImageJ anyways.
5. All tests seem to pass.

Also, I would really like if `demo_pipeline.py` was a function instead of a script, with its inputs being these large `params` dicts. Right now it's not trivial to call it from wrapper scripts and to post-process its output. I will obviously send the pull request if you deem this change useful.

### DIFF
--- a/caiman/components_evaluation.py
+++ b/caiman/components_evaluation.py
@@ -491,7 +491,7 @@ def estimate_components_quality(traces, Y, A, C, b, f, final_frate = 30, Npeaks=
             fitness_raw = np.concatenate([fitness_raw,fitness_raw__])
             fitness_delta = np.concatenate([fitness_delta, fitness_delta__])
             r_values = np.concatenate([r_values ,r_values__])
-            significant_samples = np.concatenate([significant_samples,significant_samples__])
+            # significant_samples = np.concatenate([significant_samples,significant_samples__])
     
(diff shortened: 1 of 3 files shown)


agiovann:
Excellent work!

PR #108, opened 2017-10-22
Pipeline as function

HagaiHargil:
1. Changed `demo_pipeline.py` to be a function. It has a `if __name__ == '__main__:` section to make sure that usability wasn't harmed.
2. Added some values to the parameters dictionary in preparation for subindices when reading stacks.
3. Added a batch script in which you enter a foldername and it finds recursively all files in that folder and runs the pipeline on them.
4. Commented out some print that was flooding the console.

I hope the tests pass, I had a couple of issues with them.

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1089,8 +1089,7 @@ class movie(ts.timeseries):
                 cv2.waitKey(100)
 
 
-
-def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None, var_name_hdf5 = 'mov', in_memory = False, is_behavior = False):
+def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None, var_name_hdf5='mov', in_memory=False, is_behavior=False):
     """
     load movie from file. SUpports a variety of formats. tif, hdf5, npy and memory mapped. Matlab is experimental. 
 
--- a/caiman/gui/batch_process_caiman.py
+++ b/caiman/gui/batch_process_caiman.py
@@ -1,0 +1,70 @@
+try:
+    from pathlib2 import Path  # py2.7 version
+except ModuleNotFoundError:
+    from pathlib import Path  # py3 version
+
+from demos_detailed.demo_pipeline import main_pipeline
+
+foldername = Path(r'/data/David/rat_#919_280917/1')  # Enter path to folder containing all files
+all_files = [str(filename) for filename in foldername.rglob('*.tif')]  # Filter out unwanted files in the .rglob parameter
+print("The following files were found: \n{}".format(all_files))
+
+params_movie = {'fname': all_files,  # Don't edit
+                'niter_rig': 1,
+                'max_shifts': (3, 3),  # maximum allow rigid shift
+                'splits_rig': 20,  # for parallelization split the movies in  num_splits chuncks across time
+                # if none all the splits are processed and the movie is saved
+                'num_splits_to_process_rig': None,
+                # intervals at which patches are laid out for motion correction
+                'strides': (48, 48),
+                # overlap between pathes (size of patch strides+overlaps)
+                'overlaps': (24, 24),
+                'splits_els': 28,  # for parallelization split the movies in  num_splits chuncks across time
+                # if none all the splits are processed and the movie is saved
+                'num_splits_to_process_els': [14, None],
+                'upsample_factor_grid': 4,  # upsample factor to avoid smearing when merging patches
+                # maximum deviation allowed for patch with respect to rigid
+                # shift
+                'max_deviation_rigid': 2,
+                'p': 1,  # order of the autoregressive system
+                'merge_thresh': 0.8,  # merging threshold, max correlation allowed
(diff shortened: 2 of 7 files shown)


HagaiHargil:
One more question - 
[Here](https://github.com/simonsfoundation/CaImAn/blob/master/demos_detailed/demo_pipeline.py#L196) you load the movie into memory, but a [few lines below](https://github.com/simonsfoundation/CaImAn/blob/master/demos_detailed/demo_pipeline.py#L214) you load again the entire movie into memory, but take only a part of it.

Wouldn't it be easier to slice the already-loaded movie, like that:
```python
min_mov = m_orig[:400, Ellipsis].min()
```
I think it will also increase performance, especially when dealing with stacks that are several GB in size. I can add this small fix to this PR.

HagaiHargil:
I've allowed myself to just commit these changes. I can roll back this last commit if you don't want it.

agiovann:
Hello @HagaiHargil, thank you. It is difficult for to evaluate all the changes since you also added some indenting. If you want to change indentation you should before submit a pull request for changes and then a pull request for indentation. I know it is frustrating but my editor unfortunately does not comply automatically with PEP and having it to do so is painful. Could you submit a pull request containing ONLY the changes and not fixing indentation? Thank you!

HagaiHargil:
Hi,
Is there a chance you review these changes in GitHub? I do these corrections almost automatically wherever PyCharm wants me to... :X I never realized they could be a problem, sorry.

agiovann:
@HagaiHargil I can, but it takes much longer, since I cannot focus on the real changes. It becomes a much more involved process. If you can solve it, OK, otherwise I will do it in the next days.

HagaiHargil:
OK. I hope that by tomorrow night I'll submit a new commit that reverts the indentation changes but leaves the new "content". Next time I'll resist my urge to conform to PEP8 :D

epnev:
@HagaiHargil thanks again for this contribution :) Demo pipeline is cumbersome but it serves as a very detailed demo/tutorial to exhibit the different features of the code. There are things there that are repetitive (e.g., both rigid and non-rigid motion correction) and computations that are un-necessary (e.g., compute all the metrics to assess the performance of motion correction). Having this packaged in a single function with no ability to inspect/intervene kinds of defeats the purpose. I'm not opposed into having a single function for all the analysis, but perhaps we should think about exactly what features it should have

agiovann:
@HagaiHargil @epnev my feeling is that it would be useful if it is a __main__ method in the demo and controls some flags. Meaning, we should not have another file, but merely offer the possibility to run it in batch mode and decide what to run ( like for instance only rigid motion correction and CNMF, not showing images..). Etc. Open to other possibilities

agiovann:
@HagaiHargil regarding loading the movie, it was meant for people to do some explorations, but, maybe naively, we were thinking people would adapt the script to their needs. It is possible that we are overestimating the code understanding of some researchers early in the python learning curve.

HagaiHargil:
@epnev @agiovann Unknowingly I already started working towards your proposition of a main script with flags. For example, I added the `play_movie` flag that prevents movies from being played.

Regarding code understanding - I must admit that after about a week and half of pretty much constant fiddling with the pipeline, I still don't feel comfortable with it, and I consider myself to be a decent Python programmer. To be honest, I didn't even notice the double motion correction until @epnev said so (I guess the non-rigid part is the `pwrigid` function?...). I also intended to add the subindices parameter to the loading of files, but due to the complexity of this matter I'm currently only receiving exceptions.

I think there should be an obvious "Start Here" script after you finish examining the Jupyter notebooks and want to start doing real analysis. I thought it was `demo_pipeline`, but it can definitely be something else. This script should contain everything one would want - but not more than that. I.e. no double motion correction, and without actually playing the movies.

It's super-easy to create a simple Tkinter GUI that opens up when you run `demo_pipeline`. This GUI can have two parts - in the first you choose a file for analysis, and in the second you have many _checkboxes_ that control the flow of the script (motion correction type, play movies or not) and _textboxes_ that contain the parameters values. This GUI is very similar in mind to what @outlace demonstrated in the Gitter channel.

Assuming I know which parts in `demo_pipeline` are meant to do which job - creating this GUI is no more than a day's job for me.

agiovann:
@HagaiHargil @epnev excellent suggestions and feedback. It is actually not double motion correction, I think @epnev got confused, it is actually required to create a decent template for non rigid. Here you provide some options that seem interesting to make the code more understandable. I think it would be a good idea to brainstorm about the best way to go. if you and @outlace have time, it could be productive to chat in a conference call and listen to your experiences and decide together how to best proceed, since you guys have been fiddling a lot with it. Thanks a lot again for the contribution!

outlace:
@agiovann @HagaiHargil I'd be happy to conference call.   I think it would be nice to make this have a consistent, abstracted API, which is something I'm kind of working on. I think the demo pipeline does a good job at exposing all the internals but for the end-user who just wants to plug in some files and get results out the other end, it's a little overwhelming. I'm a programmer, but even so, I don't want to have to write a lot of code to use an analysis package, I just want to be able to call the right functions with the right inputs and get what I want returned.

PR #111, opened 2017-10-29
Fix potential error in timeseries.py

outlace:
The opencv cv2 module no longer contains a cv2.cv module, so I was getting an error trying to save a movie object as an AVI file. I propose a simple try/except block to handle both the old API and the new version. Works for me.


agiovann:
Thank you!

PR #112, opened 2017-10-29
Fix potential error in visualization.py

outlace:
Line 160 tries to call the power method on a ndarray `A.power(2)` however I got the error `AttributeError: 'numpy.ndarray' object has no attribute 'power'` and indeed the numpy docs: < https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.power.html > do not suggest that `ndarray` objects have a power method. You have to do `np.power(A,2)`.   This PR changes the attempted method call to an explicit power function call in the `numpy` module.


outlace:
Actually `A.power` works if `A` is a `coo_matrix`.  For some reason when I run CNMF with patches `A` is a `coo_matrix` but when I run without patches it's an `ndarray` and so I get the aforementioned error.

outlace:
Actually a better change would be this:
```python
nA2 = np.ravel(np.power(A,2).sum(0)) if type(A) == np.ndarray else np.ravel(A.power(2).sum(0))
```
that would handle both cases if it's `ndarray` or `coo_matrix`

PR #113, opened 2017-10-30
Fix potential error in visualization.py

outlace:
Line 160 tries to call the power method on a `ndarray A.power(2)` however I got the error `AttributeError: 'numpy.ndarray' object has no attribute 'power'` and indeed the numpy docs: < https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.power.html > do not suggest that `ndarray` objects have a power method. You have to do `np.power(A,2)`. In my hands, if I run without patches mode, then the `A` matrix is a normal `ndarray` not a `coo_matrix` thus the line throws an exception. This PR will allow it to handle either case.


agiovann:
Thanks a lot @outlace

PR #120, opened 2017-11-10
Bug fixes for Windows 10 under Python 3.5.4

mingguichen:


agiovann:
Hello @mingguichen, there is something unexpected about this pull request. It says that you modified 48 files. Maybe you got caught in the middle of our update? It I am not wrong it should have only been one file. Am I correct?

mingguichen:
Yes, you are right Andrea. I modified only two files, cluster.py & spatial.py, with relatively simple edits.

agiovann:
Integrated manually

PR #126, opened 2017-11-29
Remove some dead code, also a spurious print that messes up pydoc

pgunn:
Executing code during a library import is usually a bad idea (pydoc imports libraries while parsing code, as an example). Here I removed an instance of that, making a failure to find the numba library quiet. Also removed a bit of dead code from the slurm demo pipeline.


agiovann:
Thank you Pat!

PR #142, opened 2017-12-20
Switch mac environment to conda-forge and python3

pgunn:
This passes nosetests; we should test it throughly before landing


dylex:
Does this let us get rid of the hacky for loop around nosetests on osx too?

pgunn:
I can try removing it; I'm not sure what that loop was for.

dylex:
nosetests on osx (at least that version) used some hacky way to load modules when you passed it (implicit) python files rather than module names, which broke stuff.

pgunn:
Ah, interesting. If that's still an issue, we can have conda install a specific version of nose with the features we need.

j-friedrich:
is there a reason to still use pip for tifffile and peakutils? The conda-forge versions worked fine on my Mac and it seems easier to not mix conda and pip if you don't have to.

dylex:
You should probably merge dev into your branch -- it might fix these random build failures.

PR #143, opened 2018-01-02
Add support for CUDA FFT during motion correction

neurodroid:
This patch optionally uses [`skcuda.fft`](http://scikit-cuda.readthedocs.io/en/latest/generated/skcuda.fft.fft.html) instead of OpenCV's fft or numpy's fft.

On the system used for testing (GPU: Quadro M4000; CPU: Xeon E5-2640 v4), best performance is achieved with 4 parallel processes, where the CUDA version is about as fast as the OpenCV version with 20 processes.

Further improvements could be made if the frames were not transferred one-by-one to the GPU. Instead, it would be more efficient to upload the whole movie to the GPU only once (or as much as fits into memory), and then run the CUDA FFT on chunks of GPU data. Something along these lines:
```python
# Upload movie (shape (nframes, d1, d2)) to GPU:
movie_gpu = gpuarray.to_gpu(movie)
# Get pointer to where the movie starts in GPU memory:
ptr = movie_gpu.ptr

# Run the fft on a chunk of GPU memory:
memoffset = ptr + nframe * (d1*d2) * int(frame.dtype.itemsize/8)
gpuchunk = gpuarray.GPUArray(frame.shape, frame.dtype, allocator=lambda x: memoffset)
```
However, such a strategy would require some substantial changes to the code structure, as the full preprocessed movie would need to be available before multiprocessing starts.


caiman/motion_correction.py
```diff
@@ -1552,13 +1594,50 @@ def register_translation(src_image, target_image, upsample_factor=1,
         raise NotImplementedError("Error: register_translation only supports "
                                   "subpixel registration for 2D images")
 
+    if HAS_CUDA and use_cuda:
+        from skcuda.fft import Plan
+        from skcuda.fft import fft as cudafft
+        from skcuda.fft import ifft as cudaifft
+        try:
+            cudactx
+        except:
```
pgunn: Could you use a narrower exception here?

caiman/motion_correction.py
```diff
@@ -66,6 +66,14 @@
 from cv2 import dft as fftn
 from cv2 import idft as ifftn
 opencv = True
+try:
```
pgunn: With has_cuda and use_cuda as the way this works, this means:
1) If the user doesn't have cuda installed, their code will do the right thing even if they request cuda; it'll just ignore their request
2) If the user has cuda installed and they request cuda, it will use cuda
3) If the user has cuda installed and they request cuda but it doesn't work (can't get the device, or something goes wrong initialising it because of dependency issues, or ...) then they're in for some messiness

I think this is probably a fine approach, with the alternative that comes immediately to mind to try to initialise cuda if it is requested, and if it fails fallback to the non-cuda algorithms transparently as in case 1 above.

caiman/motion_correction.py
```diff
@@ -1552,13 +1594,50 @@ def register_translation(src_image, target_image, upsample_factor=1,
         raise NotImplementedError("Error: register_translation only supports "
                                   "subpixel registration for 2D images")
 
+    if HAS_CUDA and use_cuda:
+        from skcuda.fft import Plan
```
pgunn: I think we might want an alternate environment file that has conda install this as a dependency

caiman/motion_correction.py
```diff
@@ -1552,13 +1594,50 @@ def register_translation(src_image, target_image, upsample_factor=1,
...
         src_freq = src_image
         target_freq = target_image
     # real data needs to be fft'd.
     elif space.lower() == 'real':
-        if opencv:
+        if HAS_CUDA and use_cuda:
+            # src_image_cpx = np.array(src_image, dtype=np.complex128, copy=False)
+            # target_image_cpx = np.array(target_image, dtype=np.complex128, copy=False)
+
+            image_gpu = gpuarray.to_gpu(np.stack((src_image, target_image)).astype(np.complex128))
```
pgunn: Longer-term we may want to pull this (and its opencv and other implementations) out into separate functions as this function has gotten fairly long. Not something you'd need to do though.

caiman/motion_correction.py
agiovann: I would not do it for the moment, but I would add a comment in the README alerting the users about it.

pgunn:
Neurodroid: One of the things I'm curious about with your more ambitious idea is whether the unified memory interface (cudaMallocManaged) would let us not worry about the GPU memory size (and whether the upper-level libraries use that interface yet). I feel I have a lot of reading to do about this; not sure if you already know.

I'm looking at:
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#um-unified-memory-programming-hd

caiman/motion_correction.py
neurodroid: Getting cuda to run with multiprocessing was difficult. The order of imports matters (hence the delayed import of skcuda) and the device context needs to be at global scope. I'd therefore advise against delayed initialization of cuda as it may break multiprocessing compatibility. However, there's now a message in case the user selects CUDA although it is unavailable.

caiman/motion_correction.py
neurodroid: Done.

neurodroid:
> Neurodroid: One of the things I'm curious about with your more ambitious idea is whether the unified memory interface (cudaMallocManaged) would let us not worry about the GPU memory size (and whether the upper-level libraries use that interface yet). I feel I have a lot of reading to do about this; not sure if you already know.

I'm not sufficiently familiar with CUDA programming to say anything about the advantages of the managed memory interface. [Looks like it's accessible from pycuda though](https://documen.tician.de/pycuda/driver.html#managed-memory). It may cause some additional problems when used along with multiprocessing, but then again if we load the whole movie into the gpu in a single go and then perform the ffts on large batches, we're probably better off with a single process anyway.

PR #172, opened 2018-03-07
Add cvxpy to environment.yml

vaiana:
Installing cvxpy after creating the caiman environment was causing issues with the tifffile package.  This two line change installs the cvxpy package through the conda channel omnia when the environment gets created and solves the problem.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,7 @@
 channels:
 - defaults
 - menpo
+- omnia
 dependencies:
 - future
 - numpy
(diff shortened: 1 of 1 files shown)


pgunn:
@valana The best way to install packages from conda-forge after the environment is already made is:

```conda install -n caiman -c conda-forge --override-channels cvxpy```
This will help avoid mixing conda-forge and defaults packages, which probably created the issues you observed.

As of right now it's probably best not to add xvxpy to the environment unless the codebase itself requires it; users can (if they're careful) install more things as locally needed.

Cheers,
Pat

PR #191, opened 2018-03-23
Change preferrred installation to use pip

pgunn:
This is a set of diffs that:
1) Changes the preferred way to run Caiman to not be in-place but rather to install (into your conda-env/virtualenv). The install script manages building the C++ extensions
2) So users still can comfortably edit the demos and sample code, it introduces the concept of a caiman data directory, and adds a script called caimanmanager that will populate that with the demos.
3) Nosetests are now launched from caimanmanager
4) Generally makes path handling more consistent across the codebase
5) Slightly improves underdocumentation I saw while performing the above changes.

The old (manually build my C++ extensions in place) workflow still works, provided you use caimanmanager to make a data directory (the code now looks there for things it used to look for in the codebase). You need to make a choice though, whether to build-and-run-in-place the old way or do an install.

This may have broken Jenkins. I think our Jenkins performs builds when there's a PR to a branch it cares about and not before, so creating this PR will let me see.


pgunn:
This is green and ready to merge!

PR #207, opened 2018-04-23
utils/cluster: For local/multiprocessing cluster, use forkserver method.

histed:
- numpy multithreaded BLAS/linalg do not work with multiprocessing's standard method, at least on MacOS.
See https://github.com/numpy/numpy/issues/654

### DIFF
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -406,7 +406,8 @@ def setup_cluster(backend='multiprocessing', n_processes=None, single_thread=Fal
                 raise Exception(
                     'A cluster is already runnning. Terminate with dview.terminate() if you want to restart.')
             c = None
-            dview = Pool(n_processes)
+            ctx = multiprocessing.get_context('forkserver')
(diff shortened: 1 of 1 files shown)


caiman/cluster.py
```diff
@@ -406,7 +406,8 @@ def setup_cluster(backend='multiprocessing', n_processes=None, single_thread=Fal
                 raise Exception(
                     'A cluster is already runnning. Terminate with dview.terminate() if you want to restart.')
             c = None
-            dview = Pool(n_processes)
+            ctx = multiprocessing.get_context('forkserver')
```
pgunn: Does this lock us into forkserver for these two backends? If so, was this intended?

histed: It makes forkserver the start method, instead of 'fork', for the multiprocessing backend only.   According to scipy apple believes that 'fork' is buggy and should not be used.
Perhaps this should be only changed for Mac.

PR #208, opened 2018-04-24
Feature1804/mpcontext maconly

histed:
updated to use 'forkserver' only on Mac.  Some doc updates to setup_cluster()

### DIFF
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -363,13 +363,20 @@ def stop_server(ipcluster='ipcluster', pdir=None, profile=None, dview=None):
 
 
 def setup_cluster(backend='multiprocessing', n_processes=None, single_thread=False):
-    """ If necessary, restart the pipyparallel cluster. If we have a slurm backend,
-        restart that instead.
+    """Setup and/or restart a parallel cluster.
 
     Parameters:
     ----------
     backend: str
-        'multiprocessing', 'ipyparallel', and 'SLURM'
+        'multiprocessing' [alias 'local'], 'ipyparallel', and 'SLURM'
+        ipyparallel and SLURM backends try to restart if cluster running.
+        backend='multiprocessing' raises an exception if a cluster is running.
+
+    Returns:
+    ----------
+        c: ipyparallel.Client object; only used for ipyparallel and SLURM backends, else None
+        dview: ipyparallel dview object, or for multiprocessing: Pool object
+        n_processes: number of workers in dview. None means guess at number of machine cores.
     """
     #todo: todocument
 
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -32,6 +32,7 @@ import numpy as np
 from .mmapping import load_memmap
 from multiprocessing import Pool
 import multiprocessing
+import platform
 #%%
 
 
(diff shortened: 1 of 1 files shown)


caiman/cluster.py
```diff
@@ -406,7 +407,12 @@ def setup_cluster(backend='multiprocessing', n_processes=None, single_thread=Fal
                 raise Exception(
                     'A cluster is already runnning. Terminate with dview.terminate() if you want to restart.')
             c = None
-            dview = Pool(n_processes)
+            if platform.system() == 'Darwin':
+                mp_start_method = 'forkserver'  # default 'fork' crashes multi-threaded BLAS on Mac
+            else:
+                mp_start_method = None # use default for platform
+            ctx = multiprocessing.get_context(mp_start_method)
```
pgunn: Hi, with some configurations, this get_context is illegal (which is where it's breaking the Linux part of our Jenkins setup). If you could hoist this version of it into the Darwin block and have the old call to Pool (without  the ctx) in the else block, that'd prevent breakage on those platforms.

pgunn: Actually, this is breaking CI on OSX too. For our Python2/OSX, it fails with the get_context like our Linux tests do.

For Python3/OSX, it's failing with a data error:

```  File "/Users/jenkins/workspace/CaImAn_PR-208-DB3OXXZWJBG3JQLECJIW2BRE7TXMWHMCJYICGXB2PBZQFDBHNTQQ/caiman/source_extraction/cnmf/cnmf.py", line 508, in fit
    Yr, A, b, C, f, dview=self.dview, bl=None, c1=None, sn=None, g=None, **options['temporal_params'])
  File "/Users/jenkins/workspace/CaImAn_PR-208-DB3OXXZWJBG3JQLECJIW2BRE7TXMWHMCJYICGXB2PBZQFDBHNTQQ/caiman/source_extraction/cnmf/temporal.py", line 232, in update_temporal_components
    ITER, YrA, c1, sn, g, Cin, T, nA, dview, debug, AA, kwargs)
  File "/Users/jenkins/workspace/CaImAn_PR-208-DB3OXXZWJBG3JQLECJIW2BRE7TXMWHMCJYICGXB2PBZQFDBHNTQQ/caiman/source_extraction/cnmf/temporal.py", line 371, in update_iteration
    constrained_foopsi_parallel, args_in).get(4294967)
  File "/Users/jenkins/workspace/CaImAn_PR-208-DB3OXXZWJBG3JQLECJIW2BRE7TXMWHMCJYICGXB2PBZQFDBHNTQQ/test/osx-python3/lib/python3.5/multiprocessing/pool.py", line 644, in get
    raise self._value
numpy.linalg.linalg.LinAlgError: Array must not contain infs or NaNs```

(I'm assuming you can't see our Jenkins)

**Comment by epnev** [comment]
Hi @histed 
after some further testing we are going to include this in the dev branch which we plan to merge into master within the next day or two. If you want, you can replicate your PR there. 
However, in my architecture it seems important to use the `force` option that is not callable from the `get_context` command. So we can do something along the lines of 
```
multiprocessing.set_start_method('forkserver', force=True)
```
inside an if loop

histed:
@epnev In my hands using a multiprocessing context to start the pool doesn't need force=True.  But as long as you set forkserver only on Mac I don't see any problem with doing it this way.  Thanks.

PR #221, opened 2018-05-03
Pin keras to 2.1.5

### LINKED ISSUE #218: Installation of old keras version
In some cases, during the environment creation keras version 1.0.7 is being installed by default which then cannot be run. Pinning to version 2.1.x seems to solve the problem although unclear why it's happening in the first place.

pgunn:
Resolve #218 by pinning keras

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,7 +11,7 @@ dependencies:
 - ipython
 - ipyparallel
 - jupyter
-- keras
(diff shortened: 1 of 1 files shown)


epnev:
@pgunn why in dev and not master (or both)?

pgunn:
whoops. I'll pin in master too.

PR #235, opened 2018-05-11
Dev

agiovann:


agiovann:
+ when playing movie text is not displayed since it creates troubles in some cases
+ added nndsvd method (not used, commented for now)
+ fixed stochasticity results with dendritic case

PR #236, opened 2018-05-14
Fix motion correct imread import

outlace:
Originally it attempts to import `tiffffile` and if it fails will load `skimage.external.tifffile` but below that it imports `skimage`'s `imread` function. This was causing issues with some tif files. It is highly preferable to use the `tifffile` package when possible.

### DIFF
--- a/caiman/motion_correction.py
+++ b/caiman/motion_correction.py
@@ -63,9 +63,11 @@ except:
 
 try:
     import tifffile
+    from tifffile import imread
 except:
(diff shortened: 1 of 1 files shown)


epnev:
We added some a [CONTRIBUTING](https://github.com/flatironinstitute/CaImAn/blob/master/CONTRIBUTING.md) file with PR instructions. Long story short, we prefer them against the dev branch so it's easier to keep track of all the development. Can you rebase?

PR #250, opened 2018-05-17
uniformed way tifs are loaded (imagej and not)

agiovann:
Fixed problems with loading tif from imagej. Plus unified way tifs are loaded using the Tifffile object

### DIFF
--- a/demos/general/demo_pipeline.py
+++ b/demos/general/demo_pipeline.py
@@ -94,7 +94,7 @@ rval_thr = 0.8              # space correlation threshold for accepting a compon
 cnn_thr = 0.8               # threshold for CNN based classifier
 
 #%% download the dataset if it's not present in your folder
-if fname[0] in ['Sue_2x_3000_40_-46.tif', 'demoMovieJ.tif']:
+if fname[0] in ['Sue_2x_3000_40_-46.tif', 'demoMovie.tif']:
     fname = [download_demo(fname[0])]
 
 #%% play the movie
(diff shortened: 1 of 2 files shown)


caiman/base/movies.py
```diff
@@ -1210,19 +1213,23 @@ def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None,
...
-                                   subindices.step)
-                    input_arr = imread(file_name)[subidx]
+            with tifffile.TiffFile(file_name) as tffl:                
+                if subindices is not None:
+                    if type(subindices) is list:
+                        input_arr  = tffl.asarray(key=subindices[0])[:, subindices[1], subindices[2]]
+                    else:
+                        input_arr  = tffl.asarray(key=subindices)                        
+
+#                    elif type(subindices) is range:
```
pgunn: prefer deleting to commenting-out code.

caiman/base/movies.py
```diff
@@ -1145,7 +1146,9 @@ def animate(i):
 
 
 
-def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None, var_name_hdf5 = 'mov', in_memory = False, is_behavior = False, bottom=0, top=0, left=0, right=0, channel = None):
+def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None, 
+         var_name_hdf5 = 'mov', in_memory = False, is_behavior = False, bottom=0, 
+         top=0, left=0, right=0, channel = None, outtype=np.float32):
```
pgunn: Do you have an intended use in mind for this new parameter? Right now everything is still using np.float32

PR #253, opened 2018-05-17
Saving always with bigtiff imagej

agiovann:
Adapted the code to always use bigtiff, and more efficient saving of tif files


caiman/base/timeseries.py
```diff
@@ -168,27 +172,43 @@ def save(self, file_name, to32=True, order='F'):
...
+                
+                    
+                for i in range(self.shape[0]):
+                    if i%200 == 0:
+                        print(str(i) + ' frames saved')
+                    
+                    curfr = self[i].copy()
+                    if to32 and not('float32' in str(self.dtype)):
+                         curfr = np.clip(curfr, mn, mx)                
+                         curfr = 65536 * (curfr - mn) / (mx - mn)
```
epnev: where does this come from?

epnev: float32 ranges in [-2^15 + ,2^15-1] which is 32767. But most importantly why do you multiply with this? This would explain some weird behavior I've seen

agiovann: You are right. It was there before and I did not correct it, thank you for spotting it. I assume in the past the assumption was that it was always positive. Now we now it is not the case.

On the other hand while I was writing the code I wondered whether this is necessary at all since this is floating point...

agiovann: This behavior in the past should not have affected much since we seldom save with this function

epnev: this would happen if you would try to motion correct a uint16 image that was suddenly acquiring negative values. Now things are converted into single so it's shouldn't matter but let's look into it tomorrow.

agiovann: Wait, 65535 should be OK since I make it positive. Let's chat about this tomorrow.

agiovann:
@epnev did you look at the float32 issue? Do you have a solution?

epnev:
no, but they main issue that the snippet 
```
curfr = np.clip(curfr, mn, mx)                
curfr = 65536 * (curfr - mn) / (mx - mn)
```
will change the dynamic range of the movie, so we need to understand when that happens and prevent it.

agiovann:
@epnev if you are OK with this I am going to merge

PR #254, opened 2018-05-18
Add upgrading doc

pgunn:
How does this look?

### DIFF
--- a/UPDGRADING.md
+++ b/UPDGRADING.md
@@ -1,0 +1,17 @@
+Upgrading
+=========
+
+If you already have CaImAn installed, but want to upgrade, please follow this procedure:
+
+From the conda environment you used to install CaImAn:
+* pip uninstall caiman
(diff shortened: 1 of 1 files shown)


epnev:
Can you enrich it explaining also what happens if somebody installed using the "-e" option?

pgunn:
I initially had done that, but the steps are the same.

epnev:
In my case a simple git pull appears to be working all the time. 
We should also link to this info in the readme and potentially putting it on the wiki. We're accumulating too many .md files in the main folder. I'm pretty sure very few people click on them.

pgunn:
A simple pull won't rebuild the c++ code, and won't refresh the conda environment; if neither changes that is fine.

epnev:
Closing, added to [wiki](https://github.com/flatironinstitute/CaImAn/wiki/Updating-CaImAn).

PR #269, opened 2018-05-29
method to select files with a GUI

agiovann:
This allows to select files with a GUI. You can transform this into a button.  

You need to add into your notebook/callback 
from caiman.utils.files_gui import gui_fname
filenames = gui_fname()


agiovann:
@outlace it was not meant for approval but just for merging into your branch. Should I close it then?

PR #284, opened 2018-06-18
Memmaping slice

agiovann:
adding the possibility to slice the data in all dimensions using the slice object


agiovann:
wrongly created pull request against master

PR #290, opened 2018-06-20
enabling sliced cnmf

epnev:
This is not meant to be merged at the moment.

I made a few targeted changes to enable running cnmf on arbitrarily sliced parts of the datasets.
e.g., try running demo_caiman_basic with cnm.fit(images[:1000])
Essentially instead of passing the filename of the memory mapped object to each process it passed the sliced loaded file itself. 
Before merging, we'll need to check whether this creates overhead with respect to memory and/or speed with some big files. If not some cleaning up will be needed before pushing.


epnev:
tests now pass, but we need to test for possible speed/memory overhead.

PR #305, opened 2018-07-09
Use logging module

pgunn:
Ready for review


caiman/motion_correction.py
```diff
@@ -191,28 +192,7 @@ def __init__(self, fname, min_mov=None, dview=None, max_shifts=(6, 6), niter_rig
         self.border_nan = border_nan
         self.pw_rigid = pw_rigid
         if self.use_cuda and not HAS_CUDA:
-            print("pycuda is unavailable. Falling back to default FFT.")
-
-    def motion_correct(self, template=None, save_movie=False):
```
epnev: This method seems to be missing from the PR. Can we put it back in?

PR #311, opened 2018-07-11
allows motion correction of 3d `.npy` files

AndrewMicallef:
This is a really small addition, I've just added a couple of lines to the motion_correction file that allows you to use a numpy ('.npy') file as the source.

### DIFF
--- a/caiman/motion_correction.py
+++ b/caiman/motion_correction.py
@@ -2501,7 +2501,8 @@ def motion_correction_piecewise(fname, splits, strides, overlaps, add_to_movie=0
         # T x h x w
         array = np.load(fname, memmap_mode='r')
         shape = array.shape
-
+        del array
+        
         try:
             T, d1, d2 = shape
         except:
--- a/caiman/motion_correction.py
+++ b/caiman/motion_correction.py
@@ -2426,7 +2426,7 @@ def tile_and_correct_wrapper(params):
 
     elif extension == '.sbx':  # check if sbx file
         imgs = cm.base.movies.sbxread(img_name, idxs[0], len(idxs))
-    elif extension == '.sima' or extension == '.hdf5' or extension == '.h5':
+    elif extension in ('.sima', '.hdf5', '.h5', '.npy'):
         imgs = cm.load(img_name, subindices=list(idxs))
     elif extension == '.avi':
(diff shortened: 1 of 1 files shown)


caiman/motion_correction.py
```diff
@@ -2497,7 +2497,15 @@ def motion_correction_piecewise(fname, splits, strides, overlaps, add_to_movie=0
         T = shape[0]
         del dataset
     elif extension == '.npy':
-        raise Exception('Numpy not supported at the moment')
+        # assumes the data is saved in a 3x3 array for the moment with order
```
agiovann: @AndrewMicallef we might have a problem here. You should be able to load only part of the movie in memory, otherwise when you perform parallel motion correction every process will load in memory the full movie, therefore occupying a total RAM memory n_processes*original_movie_size_single_precision.

AndrewMicallef: I was following the pattern of the other file types here, when I hacked this together it wasn't clear how the data actually get's past to the motion correction algorithm. It seams to me that at this juncture the only used information is the array shape. 
In any case, using np.load(array, memmap_mode='r') prevents the array being loaded into the memory, so I'm not sure you get the explosion you are mentioning. Either way, because the array I use doesn't go anywhere it is a bit of a waste, so I'd probably be better just grabbing the header info from the numpy file...

I'm actually heading on holiday for two weeks, so I won't be hacking a way at this for a little while

agiovann: Memory mapping the npy file is a great idea, when you are back we can chat about it. Thank you!

pgunn:
@agiovann @AndrewMicallef I'm realising we never got back to this, and I'm hoping to close out old issues; what do you think we should do with this PR?

AndrewMicallef:
@pgunn I'm happy for you to close if you would like. That 2 week holiday turned into a full life reevaluation. I havent done any image processing since, so can't speak for the demand on this.

That said @agiovann if you still want to talk about memory mapping I'm willing to take another look at this.

AndrewMicallef:
Yeah this issue can be closed

[These llines](https://github.com/flatironinstitute/CaImAn/blob/13e88d11ba53f9a9fa426f191dcc1f5ec4ca07e8/caiman/base/movies.py#L1641-L1658) in `caiman/base/movies.py : load` appear to do what I was after

PR #315, opened 2018-07-12
made online algorithm compatible with Estimates object

agiovann:


agiovann:
groundtruth generated but not pushed

PR #328, opened 2018-07-22
uniform way to write/read h5 files

epnev:
This is to deal with issue https://github.com/flatironinstitute/CaImAn/issues/326


pgunn:
If this comes up again we might like to make a map to consult of types to common extensions, although if this remains the only type that has multiple reasonable extensions, no need for that.

zbarry:
Tested this commit out myself, and it works with both extensions on my hdf5 files.

PR #359, opened 2018-08-23
Model diff: Fix docstrings for caiman/base/rois.py

pgunn:
This is a model diff demonstrating how to document functions/methods that will make documentation-related tooling happier with our codebase. "make html" in the docs dir will help find other things that sphinx is angry about, but we have a lot of code to go through to tidy.

@epnev


epnev:
@pgunn Looks good. @agiovann told me you have a script you can run through the rest of the code to do this automatically.

PR #369, opened 2018-09-20
non-distributed shape update every update_freq (instead minibatch_sha…

j-friedrich:
…pe) frames


j-friedrich:
Might still want to add the lag `minibatch_suff_stat` also to the batch update of the sufficient stats.

PR #378, opened 2018-10-03
speed up OnACID by ~15% if testing to add multiple components

j-friedrich:
more efficient computations in `get_candidate_components`
fix batch update of sufficient statistics to use `min(update_freq, minibatch_shape)`


epnev:
@j-friedrich Thanks, will check. Please don't merge.

j-friedrich:
While `loaded_model.predict` is clearly the bottleneck, the `median` in line 1423 of `online_cnmf.py` is also slow. Could we change it to mean or skip entirely cause we take only the highest `min_num_trial` peaks anyways?

PR #417, opened 2018-11-27
Change -f to --file

davclark:
According to the docs, `-f` means `--force` in `conda create` (`conda env update` unfortunately does use `-f` as `--file`). Pretty sure that's not what you mean!

Signed-off-by: dav@gigantum.com


davclark:
NM - I had a type-o on my end, I missed the `env create` instead of just `create`. Apologies!

pgunn:
No worries - some parts of the conda tooling are a bit confusing (it's unclear to me, for example, why both "conda create" and "conda env create" exist given that they both do almost the same thing).

PR #426, opened 2019-01-07
update opencv

tbenst:
per https://github.com/flatironinstitute/CaImAn/issues/425

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
 - matplotlib=2.0.2
 - nose
 - numpy
-- opencv=3.2
+- opencv=3.4
 - peakutils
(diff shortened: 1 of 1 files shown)


pgunn:
I would prefer to fix this by either merging dev to master or just cherry-picking  https://github.com/flatironinstitute/CaImAn/commit/6dea08efc4b4962d4d7b261c14c9ebb0d0656cbe into master.

pgunn:
I think we're going to do the more complete change to environment.yml, but thank you for the pull request anyhow; it's a great and appreciated thing when people are proactive in fixing problems, even though we don't always accept them.

Best wishes,
Pat

Issue #186, opened 2018-03-19
Docker container for CaImAn?


Cumol:
Have you thought about providing a docker image for CaImAn?

agiovann:
Dear @Cumol , here you have a file to generate a docker image https://github.com/flatironinstitute/CaImAn/blob/master/Dockerfile. 

we still are not distributing one, it will very likely come with the next release

Best,

Andrea

DaveMcBrave:
Hi,

is the distribution via docker hub still on the roadmap? This would be much appreated, as we want to use it on HPC.

Thanks!

epnev:
@agiovann @pgunn any updates on this?

agiovann:
@pgunn are we still using docker for testing purposes, at least in linux. If so, we just need to point to the right docker file

pgunn:
We are, but largely just as a container for our standard install process, for isolation. Docker there doesn't automate the entire thing.

The dockerfile we have in the repo is unmaintained and almost certainly doesn't work. If we wanted to provide a standard way to use Caiman with docker again (where the docker script sets things up) we'd need to start that basically from scratch.

We should chat on whether (re)supporting Docker should be on our roadmap.

paulklemm:
We would benefit quite a lot from a maintained docker file, since we can use it with singularity on our HPC cluster. This makes it easy to have every node on the cluster running identical environments, makes parallelized computing way easier.

pgunn:
@paulklemm We're chatting over future plans and docker has come up; in the meantime you may be able to extract the information you need from the Jenkinsfile in the repo (which in our setup is used to build a dockerfile on-the-fly in CI). I can't promise how soon we'll have a working Dockerfile for general use, but I hope this is at least helpful.

pgunn:
@paulklemm Hi,
I have an early Dockerfile that might do the job for you. Given that whatever you're doing in the cluster will require you to customise the dockerfile anyhow, I feel comfortable giving it to you in this very early undertested state:

```
FROM continuumio/anaconda3

RUN conda config --set always_yes yes
RUN conda update --yes conda
RUN mkdir src && cd src && git clone -b dev git@github.com:flatironinstitute/CaImAn.git && cd CaImAn && conda env create -n caiman -f environment.yml && conda activate caiman && pip install .
RUN conda activate caiman && caimanmanager.py install
```

(any feedback on this is welcome - some variant of this will make its way into dev sometime soonish. Note that this does not set environment variables that you may need to set)

[...]

PR #454, opened 2019-02-12
Provide a modern Dockerfile

pgunn:
This now successfully builds a Docker container with CaImAn inside (use the "caiman" conda environment).


pgunn:
This now builds correctly. Intended as a fix for #186

PR #456, opened 2019-02-17
Fix threshold plot contours

agiovann:
This PR tries to fix issue with thresholds in plot contours. One should actually modify the original contours function to accept a single parameter method and a single parameter threshold instead of having one parameter for each threshold type (maxthr and nrgthr). However, this might cause back compatibility issues, so I did not touch it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,41 +1,9 @@
-FROM ubuntu
+FROM continuumio/anaconda3
 
-RUN apt-get update
-RUN apt-get install -y libglib2.0-0
-RUN apt-get install -y git wget
-RUN apt-get install bzip2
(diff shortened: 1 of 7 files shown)


epnev:
@agiovann I think this is unnecessary. The function `get_contours` will compute the coordinates with a given threshold and method it will then just plot them without re-computing them. If you want to change the threshold then you'll need to first set `cnm.estimates.coordinates=None`. The `get_contours` function takes as an input both threshold method and the parameter `thr` that specifies the level.

epnev:
You can argue about setting a flag that re-computes the coordinates instead of re-using the old ones, but this can be done more efficiently.

PR #458, opened 2019-02-18
workaround to urllib permission error

agiovann:
Added a workaround to the permission error. This should solve issue #457. However, we might want to find a more permanent solution storage that would not block the urlreq. I did not propagate the changes to master since I do not know if you are OK with this patch.

### DIFF
--- a/caiman/utils/utils.py
+++ b/caiman/utils/utils.py
@@ -82,7 +82,15 @@ def download_demo(name='Sue_2x_3000_40_-46.tif', save_folder=''):
             url = file_dict[name]
             logging.info(f"downloading {name} with urllib")
             logging.info(f"GET {url} HTTP/1.1")
-            f = urlopen(url)
+            try:
+                f = urlopen(url)
+            except:
+                logging.info(f"Trying to set user agent to download demo")
+                from urllib.request import Request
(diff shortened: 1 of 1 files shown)


agiovann:
Guys, please try to process ASAP since many people are trying to use the software right now.

caiman/utils/utils.py
```diff
@@ -82,7 +82,15 @@ def download_demo(name='Sue_2x_3000_40_-46.tif', save_folder=''):
             url = file_dict[name]
             logging.info(f"downloading {name} with urllib")
             logging.info(f"GET {url} HTTP/1.1")
-            f = urlopen(url)
+            try:
+                f = urlopen(url)
+            except:
+                logging.info(f"Trying to set user agent to download demo")
```
epnev: @agiovann typo here?

agiovann: Is bad english the problem?

epnev: There is an "f" before the quotes and the tests are failing

agiovann: I coped from the lines 83-84 above the syntax. Not sure why it is failing... works for me

pgunn: The f should not be an issue - that's a standard python feature called an f-string.

PR #462, opened 2019-02-19
Change domain for datafiles hosting to a direct (non-Cloudfront) endpoint

pgunn:
This is meant as a better fix for #457 . The old endpoint was pointing at a CloudFront cache which has anti-abuse features that sometimes considered API-like web requests invalid and returned 500. Dylan set up a direct DNS name for us that points right at our users service (but otherwise has identical semantics to users.flatironinstitute.org) so that won't be a concern anymore.

### DIFF
--- a/caiman/utils/utils.py
+++ b/caiman/utils/utils.py
@@ -65,14 +65,14 @@ def download_demo(name='Sue_2x_3000_40_-46.tif', save_folder=''):
     #\bug
     #\warning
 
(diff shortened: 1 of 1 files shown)


epnev:
@pgunn Where are the files hosted now? Downloading is significantly slower.

pgunn:
They're served directly on our servers without the Cloudfront cacheing layer.

PR #473, opened 2019-02-25
Speed up online

agiovann:
This small changes adds a good speedup to online. I also added another way of spelling the .tif extension (.TIFF) 

Not sure what happened in terms of merging, it included some other commits from Pat, probably because I merged without committing (my bad). @pgunn can you check that everything is OK? the only real changes here are 

- added an equivalent extinsion for tif files
- speed up online by ordering in the right directions array

Let me know!

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1146,7 +1146,8 @@ def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None,
 
         return load_movie_chain(file_name,fr=fr, start_time=start_time,
                      meta_data=meta_data, subindices=subindices,
-                     bottom=bottom, top=top, left=left, right=right, channel = channel, outtype=outtype)
+                     bottom=bottom, top=top, left=left, right=right, 
+                     channel = channel, outtype=outtype)
 
     if bottom != 0:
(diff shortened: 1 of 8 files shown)


pgunn:
The diff looked weird because you based it off dev but were doing the PR against master. I just adjusted it to be a PR against dev (assuming that's what you wanted).

caiman/base/movies.py
```diff
@@ -1157,7 +1158,7 @@ def load(file_name,fr=30,start_time=0,meta_data=None,subindices=None,shape=None,
     if os.path.exists(file_name):
         _, extension = os.path.splitext(file_name)[:2]
 
-        if extension == '.tif' or extension == '.tiff':  # load avi file
+        if extension == '.tif' or extension == '.tiff' or extension == '.TIF':  # load avi file
```
pgunn: We should probably just lowercase the extension right after splitting it off from the file before comparison, so we don't need to introduce case variants for all the extensions.

agiovann:
Oops sorry. Rookie mistake. Sure.

caiman/base/movies.py
epnev: Agree. I looked for the appearances of the word `extension` and switched it to always be lowercase. Hopefully that should capture everything

PR #483, opened 2019-03-06
trying out python 3.7

epnev:
do not merge


epnev:
Installation failed. It appears that tensorflow compatible with python 3.7 only in version 1.13 which is not being distributed by conda yet.

PR #489, opened 2019-03-19
Second batch of type annotations

pgunn:
A) focused work on making utils/utils.py complete
B) less focused work on poking at other files to do the minimum needed to squelch warnings from mypy to lessen noise


caiman/base/movies.py
```diff
@@ -108,6 +109,7 @@ def __new__(cls, input_arr, **kwargs):
 
     def motion_correction_online(self, max_shift_w=25, max_shift_h=25, init_frames_template=100,
                                  show_movie=False, bilateral_blur=False, template=None, min_count=1000):
+        # FIXME: This does not match the signature in motion_correction.py:motion_correct_online
```
pgunn: Almost certainly a bug

caiman/base/timeseries.py
```diff
@@ -253,7 +253,6 @@ def concatenate(*args, **kwargs):
     """
     # todo: todocument return
 
-    obj = []
```
pgunn: Initially having this be a list didn't serve a point and resulted in it changing types in the function body. Removing this makes the type consistent across the function.

caiman/components_evaluation.py
```diff
@@ -414,20 +415,11 @@ def evaluate_components(Y, traces, A, C, b, f, final_frate, remove_baseline=True
 
 
 #%%
-# FIXME xrange is python2-specific
```
pgunn: This function wasn't used anywhere. Fixing it would've been easy for python3, but no point if nobody's using it.

caiman/components_evaluation.py
```diff
@@ -414,20 +415,11 @@ def evaluate_components(Y, traces, A, C, b, f, final_frate, remove_baseline=True
...
-
 
 def grouper(n, iterable, fillvalue=None):
     "grouper(3, 'ABCDEFG', 'x') --> ABC DEF Gxx"
     args = [iter(iterable)] * n
-    try:  # py3
-        return itertools.zip_longest(*args, fillvalue=fillvalue)
-    except:  # py2
-        return itertools.izip_longest(*args, fillvalue=fillvalue)
+    return itertools.zip_longest(*args, fillvalue=fillvalue)
```
pgunn: python3 for the win

caiman/motion_correction.py
```diff
@@ -768,7 +769,7 @@ def motion_correct_online(movie_iterable, add_to_movie, max_shift_w=25, max_shif
                 big_mov[:, idx_frame] = np.reshape(
                     new_img, np.prod(dims[1:]), order='F')
 
-            if return_mov and (n_iter == (n + 1)):
+            if mov is not None and (n_iter == (n + 1)):
```
pgunn: This should not change functionality, and makes it clear to the type inference engine that mov has a value for this block

caiman/motion_correction.py
```diff
@@ -2557,8 +2558,8 @@ def motion_correction_piecewise(fname, splits, strides, overlaps, add_to_movie=0
     if save_movie:
         if base_name is None:
             base_name = os.path.split(fname)[1][:-4]
-        fname_tot = base_name + '_d1_' + str(dims[0]) + '_d2_' + str(dims[1]) + '_d3_' + str(
-            1 if len(dims) == 2 else dims[2]) + '_order_' + str(order) + '_frames_' + str(T) + '_.mmap'
+        fname_tot = base_name + '_d1_' + str(dims[0]) + '_d2_' + str(dims[1]) + '_d3_1' + \
```
pgunn: Stared carefully at the preceding code and convinced myself that dims never has a third dimension here.

Longer term we may want to make utility functions that handle constructing names for memmap files (possibly with a default path).

caiman/utils/utils.py
```diff
@@ -21,31 +21,32 @@
 #\pre none
 
 import cv2
-import logging
 import h5py
-
+import logging
+import matplotlib.pyplot as plt
```
pgunn: This was dynamically imported later. Changed to a static import.

caiman/utils/utils.py
```diff
@@ -137,15 +139,15 @@ def si_parse(imd):
 
     """
 
-    imd = imd.split('\n')
-    imd = [i for i in imd if '=' in i]
-    imd = [i.split('=') for i in imd]
-    imd = [[ii.strip(' \r') for ii in i] for i in imd]
-    imd = {i[0]: val_parse(i[1]) for i in imd}
-    return imd
+    imddata:Any = imd.split('\n')
```
pgunn: Using a different variable lets us have the input be restricted to be a string type

caiman/utils/utils.py
```diff
@@ -264,23 +264,23 @@ def gen_data(dims=(48, 48), N=10, sig=(3, 3), tau=1., noise=.3, T=2000,
         plt.imshow(Cn, cmap=cmap)
         plt.title('Correlation')
         plt.show()
-    return Yr, trueC, trueS, trueA, trueb, truef, centers, dims
+    return Yr, trueC, trueS, trueA, trueb, truef, centers, dims # XXX dims is always the same as passed into the function?
```
pgunn: Unclear why we pass dims back if it can never change

caiman/base/movies.py
epnev: Not sure I understand. Where is the bug?

caiman/motion_correction.py
epnev: @pgunn This might change in the future if we enable 3d motion correction. Can you revert?

caiman/base/movies.py
pgunn: Sorry, I marked the wrong line; should be the line below where it calls motion_correct_online (but doesn't call it correctly)

epnev: can you specify what the error is? Seems to me poor coding that just replicates a function rather than a bug

pgunn: The positional parameters to motion_correct_online() are:
movie_iterable, add_to_movie
(followed by a number of named arguments)

the call below only provides one positional parameter (the self object).

PR #491, opened 2019-03-20
Update Dockerfile

maxtue:
Adding apt-get update since certain packages could not be fetched without it  which lead to abortion of setup.


pgunn:
@maxtue looks good, thanks. I changed the PR to be against dev rather than master, as we prefer PRs land in dev first.

maxtue:
Geht klar, das wusste ich nicht. :) Thanks!

pgunn:
Alles gut, danke für der Patch.

PR #504, opened 2019-04-03
Type annotation effort 2019-04-03

pgunn:
This:
* adds more type annotations
* corrects the spelling of the word 'indices' across the codebase
* Avoids use of 'dict' as a variable name (as that's also a type, so it's confusing)
* removes a few functions that were not being used anywhere (and other cruft)
* has some very minor refactoring
* Adds some TODOs to think about for later work
* makes docs a little better in some places

### DIFF
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -34,27 +34,13 @@ import shutil
 import subprocess
 import sys
 import time
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from .mmapping import load_memmap
 
 logger = logging.getLogger(__name__)
 
-def get_patches_from_image(img, shapes, overlaps):
-    # todo todocument
-    d1, d2 = np.shape(img)
-    rf = np.divide(shapes, 2)
-    _, coords_2d = extract_patch_coordinates(d1, d2, rf=rf, stride=overlaps)
-    imgs = np.empty(coords_2d.shape[:2], dtype=np.object)
-
-    for idx_0, count_0 in enumerate(coords_2d):
-        for idx_1, count_1 in enumerate(count_0):
-            imgs[idx_0, idx_1] = img[count_1[0], count_1[1]]
-
-    return imgs, coords_2d
-#%%
-
-
-def extract_patch_coordinates(dims, rf, stride, border_pix=0, indeces=[slice(None)]*2):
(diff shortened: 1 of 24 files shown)


caiman/cluster.py
```diff
@@ -34,27 +34,13 @@
 import subprocess
 import sys
 import time
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from .mmapping import load_memmap
 
 logger = logging.getLogger(__name__)
 
-def get_patches_from_image(img, shapes, overlaps):
```
epnev: @pgunn Why is this method being removed?

caiman/components_evaluation.py
```diff
@@ -29,52 +29,14 @@
 except:
     pass
 
-def estimate_noise_mode(traces, robust_std=False, use_mode_fast=False, return_all=False):
```
epnev: @pgunn This is also being removed?

caiman/components_evaluation.py
```diff
@@ -291,8 +249,8 @@ def evaluate_components_CNN(A, dims, gSig, model_name=os.path.join(caiman_datadi
 #%%
 
 
-def evaluate_components(Y, traces, A, C, b, f, final_frate, remove_baseline=True, N=5, robust_std=False,
-                        Athresh=0.1, Npeaks=5, thresh_C=0.3, sigma_factor=3.):
+def evaluate_components(Y:np.ndarray, traces:np.ndarray, A, C, b, f, final_frate, remove_baseline:bool=True, N:int=5, robust_std:bool=False,
```
epnev: @pgunn I think that Y can also be a memory mapped array here.

caiman/cluster.py
pgunn: Was unused. If that's not a reason to remove it I can revert this part. (same as below)

epnev:
@pgunn Can you move the unused function into a separate file? Perhaps start an unused.py file in the utils folder. I can see both of these functions being useful at some point, so wouldn't want to completely get rid of them at this point. Once you do that and do a further rebase with the current dev I can merge.

pgunn:
> @pgunn Can you move the unused function into a separate file? Perhaps start an unused.py file in the utils folder. I can see both of these functions being useful at some point, so wouldn't want to completely get rid of them at this point. Once you do that and do a further rebase with the current dev I can merge.

I made an unused.py as requested.

PR #517, opened 2019-04-24
NaN check in background component decomposition

rbedi:
@epnev There was one additional place where there NaNs were causing problems -- in the NMF on the background temporal components. Filtering out components with NaNs fixes this problem, which I think is OK since these components aren't meaningfully contributing to background anyways.


epnev:
This looks good, thanks!

Issue #519, opened 2019-04-24
`setup_cluster` check for existing cluster precludes using other multiprocessing libraries


zbarry:
For example, using a Dask `LocalCluster`:

```python
import caiman as cm
from dask.distributed import Client, LocalCluster

client = Client(LocalCluster(n_workers=12))

c, dview, n_processes = cm.cluster.setup_cluster(
    backend='local',
    n_processes=20,
)
```

```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-31-0650e205d9c4> in <module>
      1 c, dview, n_processes = cm.cluster.setup_cluster(
      2         backend='local',
----> 3         n_processes=num_processes,
      4     )

/data/external_repos/CaImAn/caiman/cluster.py in setup_cluster(backend, n_processes, single_thread)
    409             if len(multiprocessing.active_children()) > 0:
    410                 raise Exception(
--> 411                     'A cluster is already runnning. Terminate with dview.terminate() if you want to restart.')
    412             if (platform.system() == 'Darwin') and (sys.version_info > (3, 0)):
    413                 try:

Exception: A cluster is already runnning. Terminate with dview.terminate() if you want to restart.
```

Do you think there might be a more sophisticated way to check if there is a CaImAn multiprocessing pool started vs. just checking if any pool exists?

pgunn:
There may be, although I'm worried that anyone moving in this direction would find code issues that we wouldn't be able to help them with (subtle bugs that come from untested interactions, things like that). Some time back we had some issues with OpenCV, MLK, and other pieces of software and their interactions with clustering libraries - getting to the bottom of them took a fair bit of time.

zbarry:
What about going the other way - some flag for the user to specify to ignore this check and have them deal with the consequences of that decision?

pgunn:
I don't think that's too bad (would want to see what Eftychios thinks though); if they bypass it then they're also taking on the risk of having many clusters running accidentally.

zbarry:
Yeah, exactly. It would be something you'd note in the docstring - "use at your own risk because <reasons>" but not necessarily even show as a possibility in the example notebooks.

epnev:
@zbarry I'm fine with a flag that prevents this by default and prints a message if the user chooses to do otherwise. If you want create a PR to our dev branch.

zbarry:
@epnev you want it to be `logging.INFO` or `WARN`?

pgunn:
[...]

PR #523, opened 2019-04-25
Setup cluster ignore flag

zbarry:
Addresses #519 

@pgunn , @epnev 

Whitespace removal was PyCharm's decision (with my current settings), though I kept that in there, because it's harmless.


pgunn:
This looks good to me

epnev:
thanks!

PR #531, opened 2019-05-08
 Fix bug with scaling in notebook

maxtue:
With both size dimensions being variable the image enlargened uncontrollably when using the slider of neurons for some reason. Fixing the width and varying the height will make sure that the image on the left and spikes on the right are both visible without having to scroll left or right and at the same time keep the right aspect ratio. Also d1 and d2 seem to be misnamed as d2 is constantly used for width.

### DIFF
--- a/caiman/utils/visualization.py
+++ b/caiman/utils/visualization.py
@@ -142,7 +142,7 @@ def nb_view_patches(Yr, A, C, b, f, d1, d2, YrA=None, image_neurons=None, thr=0.
             outputs of matrix factorization algorithm
 
         d1,d2: floats
-            dimensions of movie (x and y)
+            dimensions of movie (y and x)
 
         YrA:   np.ndarray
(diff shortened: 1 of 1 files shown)


epnev:
@maxtue thanks for your note. However, I don't see how the existing the code gives "uncontrollably enlarged images" given that it is guaranteed that both height and width will be at most 300 pixels. The modification you suggested might lead to very high images in FOVs with high height to width ratio. Is there something I'm missing?

maxtue:
Ok, I only had this problem on Ubuntu 18.04, no problem on a Mac I tried today so maybe not related to CaImAn. We also didn't understand what was going on, since everything else works fine in our installation and the implemented code seems completely logical as you pointed out. If nobody else is experiencing this, I close the request.

epnev:
Thanks, I'll keep it in mind if this issue appears in the future again.

PR #532, opened 2019-05-08
Zero pixels issue

davegrays:
This is a possible solution to the problem of when videos have pixels that are set to 0 across time, such as when a mask is applied to all frames. Removes components with nans and also skips pixels that are 0.


epnev:
looks good, thanks!

PR #534, opened 2019-05-15
2019 05 type annotations and cleanup

pgunn:
This contains more work on type annotations, a few docs improvements and other cleanups.

I recently learned that you can in fact reference a return type of the class you're in the middle of defining by enclosing the type name in quotes - listing it as a string defers the evaluation until the block defining the class finishes. caiman/base/movies.py has this on line 113 (in this diff), will be doing more of it in future diffs.


pgunn:
This is ready for review

pgunn:
Notice that this removes the kwargs from the call to threshold_components()

epnev:
Thanks!

> Notice that this removes the kwargs from the call to threshold_components()

There is a chance we might need this (for better or for worse)

pgunn:
> Thanks!
> 
> > Notice that this removes the kwargs from the call to threshold_components()
> 
> There is a chance we might need this (for better or for worse)

If we can take it out for now and only add it back in if someone can make a very strong case that there's no other way to do what they need done, that'd be a win - this kind of calling makes it really hard to follow code.

epnev:
I think the scripts run during the nightly build will use that script and show whether it's needed or not. So we can wait until those are up or try manually and see if this fails.

PR #535, opened 2019-05-16
New CaImAn GUI

agiovann:
New prototype interface for CaImAn:

* Interface working for the three demos demo_basic.py demo_pipeline.py and demo_OnAcid_mesoscope.py. 
* Interface can save after cleaning up components and reload
* added save methods to the three demos
* added instructions in Readme.md


epnev:
@agiovann excellent! can you rebase it against dev?

epnev:
check #536 instead.

PR #537, opened 2019-05-16
Unstable! New NWB support for saving movie and estimates

agiovann:
Unstable! New methods to save in NWB format movies and estimates. Needs to be tested.


epnev:
check #539 instead.

PR #539, opened 2019-05-16
Gui ag 2.0

agiovann:
Added support for NWB and GUI. 

* created a demo for NWB, along with methods for saving movie, loading movie, saving results (estimates)
* pyqtgraph based visual interface to validate the results of analysis
* Modified three demos to produce an hdf5 file that can be used with visual interface


caiman/base/timeseries.py
```diff
@@ -18,7 +18,14 @@
 from scipy.io import savemat
 import tifffile
 import warnings
-
+from datetime import datetime
```
pgunn: Could we keep the non-internal exports in order?

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+             location='somewhere in the brain',
+             orig_file_format='tiff'):
+        """save object in hdf5 file format
+
+        Args:
+            filename: str
+                path to the hdf5 file containing the saved object
+        """
+    #%%
+        from pynwb import NWBHDF5IO
```
pgunn: Since we're using conda, do imports inside the function make sense? Might be better to import at the top of the file unless we have a really good reason not to

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+        from pynwb.ophys import ImageSegmentation, Fluorescence,MotionCorrection
+        import os
+        
+    #%%
+        if '.nwb' != filename[-4:]:
+            raise Exception("Wrong filename")
```
pgunn: "bad extension"

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+        
+        if not os.path.isfile(filename): # if the file doesn't exist create new and add the orginal data path
+            raise Exception('filename should be an existing NWB file. \
+                            Consdier using cnmf.movie.save method to create one.')
+        
+        else: # if the file already exist in the .nwb format then just add the results to it
```
pgunn: exist -> exists

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+            pdb.set_trace()
+            with  NWBHDF5IO(filename, 'r+') as io:
+                nwbfile = io.read()
+                # Add processing results
+                mod = nwbfile.create_processing_module('Estimates', 'contains caiman estimates for the main imagin plane')
+                img_seg = ImageSegmentation()
+                mod.add_data_interface(img_seg)
+                fl = Fluorescence()
+                mod.add_data_interface(fl)
+    #            mot_crct = MotionCorrection()
```
pgunn: let's not add code that's already commented out

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -234,6 +237,10 @@ def _prepare_object(self, Yr, T, new_dims=None, idx_components=None):
         self.loaded_model = loaded_model
         return self
 
+    def load_CNMF_from_file(path):
```
pgunn: why are we adding an empty function?

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -659,6 +665,99 @@ def save(self,filename):
...
+        # Neurons
+        rrs1 = fl.create_roi_response_series('ROI_Fluorescence_Response', self.estimates.C.T, 'lumens', rt_region_roi, timestamps=timestamps)
+        # Background
+        rrs2 = fl.create_roi_response_series('Background_Fluorescence_Response', self.estimates.f.T, 'lumens', rt_region_bg, timestamps=timestamps)
+        
+        
+        with NWBHDF5IO(filename, 'w') as io:
+            io.write(nwbfile)
+
+
```
pgunn: we don't need this many spaces. 2 should do.

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -795,9 +906,9 @@ def fit_online(self, **kwargs):
         self.estimates.bl = [osi.b for osi in self.estimates.OASISinstances] if hasattr(
             self, 'OASISinstances') else [0] * self.estimates.C.shape[0]
         if self.params.get('online', 'save_online_movie'):
-            out.release()
+            out.release()
```
pgunn: why are we adding these extra spaces at the end of the line here and two lines below?

demos/general/demo_pipeline_nwb.py
```diff
@@ -0,0 +1,286 @@
...
+    pass  # For compatibility between running under Spyder and the CLI
+
+#%% Select file(s) to be processed (download if not present)
+    fnames = ['/Users/agiovann/caiman_data/example_movies/Sue_2x_3000_40_-46.nwb']  # filename to be created or processed
+    # dataset dependent parameters
+    fr = 15  # imaging rate in frames per second
+    decay_time = 0.4  # length of a typical transient in seconds
+
+
+#%% if it does not exists, create a NWB file
```
pgunn: exists -> exist

caiman/source_extraction/cnmf/estimates.py
epnev: pywnb is not a dependency and I don't want to make it one right now. We could keep the import inside a function if that means that it would not try to execute unless we call the function. Is this the case?

README.md
```diff
@@ -151,6 +153,11 @@ A visual interface will appear, you will be able to:
 
 <img src="docs/img/GUI_img.png" width="1000" align="center">
 
+## Exporting Results
+Currently it is possible to store the results in two formats:
+1. HDF5 file with all the details of the CNMF function. The results are stored in the `estimates` field.
+2. NWB file with a Processing module called `Estimates` that contains the processing results.
```
epnev: There is a lower case and an upper case `estimates` here. Let's stick to one.

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+            filename: str
+                path to the hdf5 file containing the saved object
+        """
+    #%%
+        from pynwb import NWBHDF5IO
+        from pynwb.ophys import ImageSegmentation, Fluorescence,MotionCorrection
+        import os
+        
+    #%%
+        if '.nwb' != filename[-4:]:
```
epnev: better to use `os.path.split` the file and get the extension.

demos/general/demo_pipeline_nwb.py
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env python
```
epnev: This demo does not add much compared to other demos. Better modify existing demos to include a save function where NWB is an option.

epnev:
@pgunn Is the a way to run @agiovann 's GUI from a Python editor or a Jupyter notebook?

caiman/source_extraction/cnmf/estimates.py
pgunn: It was added to environment.yml so it was made into a dependency.

pgunn:
> @pgunn Is the a way to run @agiovann 's GUI from a Python editor or a Jupyter notebook?

Will test

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1154,6 +1154,104 @@ def masks_2_neurofinder(self, dataset_name):
...
+            raise Exception("Wrong filename")
+        
+        if not os.path.isfile(filename): # if the file doesn't exist create new and add the orginal data path
+            raise Exception('filename should be an existing NWB file. \
+                            Consdier using cnmf.movie.save method to create one.')
+        
+        else: # if the file already exist in the .nwb format then just add the results to it
+            print('Saving the results...')
+            import pdb
+            pdb.set_trace()
```
epnev: remove the debugger

epnev:
subset of #559

PR #541, opened 2019-05-23
Parallel merging

epnev:
Speeding up CaImAn batch algorithm using:
- Parallel merging of components
- Determining block size automatically when calling `parallel_dot_product`


epnev:
@agiovann This is up your alley. Take a look when you get a chance.

PR #547, opened 2019-05-30
gitignore updates

histed:
I'm pushing some small changes we have maintained.  Feel free to accept/reject as desired

This change: update .gitignore


pgunn:
Hi,
I don't think we need the orig version of the file. Also, could you pull the oasis.cpp back out of this diff? I think it somehow snuck in.

histed:
Corrected, thanks.

PR #558, opened 2019-06-06
support for NWB and GUI: second attempt

agiovann:
- Incorporated support for NWB and a demo to use it
- Incorporate demos examples, save methods and documentation


epnev:
subset of #559

PR #562, opened 2019-06-11
Sbx support

agiovann:
added support for slicing sbx files (scanbox). Now can motion correct directly from the sbx file


epnev:
Nice!

PR #564, opened 2019-06-17
Run OnAcid without sniper_mode

hawkwings:
Latest update for supporting different classifiers breaks if sniper_mode is False, due to self.loaded_model, self.tf_in and self.tf_out. Fixed by setting those to None for sniper_mode=False.


epnev:
Thanks!

PR #568, opened 2019-06-25
renaming for nwb best practices

bendichter:
renames variables according to the new [NWB Best Practices guide](https://neurodatawithoutborders.github.io/best_practices)


epnev:
Thanks!

PR #569, opened 2019-06-25
Dev

epnev:
- Support for NWB files. You can now load an NWB file and save the analysis results back on the same file. Check `use_cases/NWB/demo_pipeline_NWB.py` for an example
- GUI for inspecting results, modifying quality thresholds, and manually removing components.


README.md
```diff
@@ -133,6 +135,35 @@ Alternative environments:
...
+* demo_caiman_basic.py
+* demo_pipeline.py
+
+Then you can start the visual interface by running the following command from the base caiman folder (make sure you are within your caiman environment):
+```
+ipython caiman/gui/gui_pyqtgraph_layout.py
+```
+You will then be asked to load the file that is generated at the end of each file (ending in .hdf5)
+
+A visual interface will appear, you will be able to:
```
pgunn: A GUI

README.md
```diff
@@ -133,6 +135,35 @@ Alternative environments:
...
+* regulate gain and contrast for the background image (correlation image)
+* regulate the threshold over the spatial masks to visualize components contours
+* click on neurons and see the corresponding trace and mask
+* select subset of neurons based on different quality metrics
+* save the resulting selection to a file in hdf5 format
+
+<img src="docs/img/GUI_img.png" width="1000" align="center">
+
+## Exporting results
+It is currently possible to export the results in two formats:
```
pgunn: It is possible

caiman/base/movies.py
```diff
@@ -1047,7 +1058,17 @@ def animate(i):
 
         looping = True
         terminated = False
-
+        if save_movie:
+            #fourcc = cv2.VideoWriter_fourcc('8', 'B', 'P', 'S')
```
pgunn: Why are we checking in commented-out code?

caiman/base/movies.py
```diff
@@ -1324,20 +1351,28 @@ def rgb2gray(rgb):
                     input_arr = np.vstack(input_arr)
 
             else:
+#                if extension == '.nwb':
```
pgunn: Can we remove this commented-out code from the PR?

caiman/base/timeseries.py
```diff
@@ -18,7 +18,11 @@
 from scipy.io import savemat
 import tifffile
 import warnings
-
+from datetime import datetime
```
pgunn: Can we keep the non-caiman imports in alphabetical order by library name?

caiman/base/timeseries.py
```diff
@@ -114,10 +118,31 @@ def __array_finalize__(self, obj):
...
+             var_name_hdf5='mov',
+             sess_desc='some_description',
+             identifier='some identifier',
+             exp_desc='experiment description',
+             imaging_plane_description='some imaging plane description',
+             emission_lambda=520.0,
+             indicator='OGB-1',
+             location='brain',
+             starting_time=0.,
+             experimenter='Dr Who',
```
pgunn: ^_^

caiman/gui/gui_pyqtgraph_layout.py
```diff
@@ -0,0 +1,323 @@
...
+import numpy as np
+import caiman as cm
+from caiman.source_extraction.cnmf.cnmf import load_CNMF
+from caiman.source_extraction.cnmf.online_cnmf import load_OnlineCNMF
+
+import cv2
+import scipy
+import os
+# Always start by initializing Qt (only once per application)
+app = QtGui.QApplication([])
```
pgunn: This file has a lot of code scattered about not living in functions that will run on import. Might be better to move all that to an initialisation function that people need to call.

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -509,8 +517,16 @@ def play_movie(self, imgs, q_max=99.75, q_min=2, gain_res=1,
 
         mov = caiman.concatenate((imgs[frame_range] - (not include_bck) * B,
                                   Y_rec + include_bck * B, Y_res * gain_res), axis=2)
+        
+
         if thr > 0:
-            import cv2
+            if save_movie:
+                import cv2
```
pgunn: Should probably be imported at the head of the file instead

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -509,8 +517,16 @@ def play_movie(self, imgs, q_max=99.75, q_min=2, gain_res=1,
...
+                #fourcc = cv2.VideoWriter_fourcc('8', 'B', 'P', 'S')
```
pgunn: More commented-out code

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -214,8 +220,12 @@ def _prepare_object(self, Yr, T, new_dims=None, idx_components=None):
         # setup per patch classifier
 
         if self.params.get('online', 'path_to_model') is None or self.params.get('online', 'sniper_mode') is False:
-            loaded_model = None
+            self.loaded_model = None
```
pgunn: Wondering if this should be in the initialiser instead

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -239,16 +249,20 @@ def _prepare_object(self, Yr, T, new_dims=None, idx_components=None):
...
                 model_path = '.'.join(path + ['h5', 'pb'])
                 loaded_model = load_graph(model_path)
                 self.tf_in = loaded_model.get_tensor_by_name('prefix/conv2d_1_input:0')
                 self.tf_out = loaded_model.get_tensor_by_name('prefix/output_node0:0')
-                self.loaded_model = tf.Session(graph=loaded_model)
+                loaded_model = tf.Session(graph=loaded_model)
+        self.loaded_model = loaded_model
         return self
 
+    def load_CNMF_from_file(path):
```
pgunn: ?

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -676,6 +694,88 @@ def save(self,filename):
         else:
             raise Exception("Unsupported file extension")
 
+#    def save_NWB(self, filename, ID, sess_desc='CaImAn Results', exp_desc=None,
```
pgunn: Why are we adding this commented-out function? (it also seems to import a lot of stuff at runtime that if we were to uncomment it should be imported at include time instead)

caiman/utils/utils.py
```diff
@@ -353,7 +353,7 @@ def cell_magic_wand_wrapper(params):
 #%% From https://codereview.stackexchange.com/questions/120802/recursively-save-python-dictionaries-to-hdf5-files-using-h5py
 
 
-def save_dict_to_hdf5(dic:Dict, filename:str) -> None:
+def save_dict_to_hdf5(dic:Dict, filename:str, subdir:str='/') -> None:
```
pgunn: Wondering if subdir is necessary - os.path.join() does the right thing in a cross-platform way

caiman/base/movies.py
epnev: @pgunn There is a bunch of different codecs for saving a video through openCV. Each will work in some platforms, but not in others. It would actually be pretty helpful if you could spend some time investigating this, and come up with some rules about which codec to use when.

caiman/utils/utils.py
epnev: @pgunn There might be an issue there, although I just wrote a test for it that passed. Take a look?

pgunn: Actually, I'm totally wrong on this - I misread what it's doing. These are paths *inside* the hdf5 file, not real filesystem paths. os.path.join() is for filesystem paths.

PR #577, opened 2019-07-12
gui updated

libbyli33:
neurons/background/reset mode, scrolling effect, trace plotted

### DIFF
--- a/caiman/gui/gui_pyqtgraph_layout.py
+++ b/caiman/gui/gui_pyqtgraph_layout.py
@@ -61,15 +61,12 @@ Cn = cnm_obj.estimates.Cn
 estimates = cnm_obj.estimates
 min_mov_denoise = np.min(estimates.A.dot(estimates.C))
 max_mov_denoise = np.max(estimates.A.dot(estimates.C))
-min_background = []
-max_background = []
 background_num = -1
 neuron_selected = False
 nr_index = 0
 
-for i in range(0,estimates.f.shape[0]):
-    min_background.append(np.min(estimates.b[:,i].reshape(4800,1).dot(estimates.f[i,:].reshape(1,2000))))
-    max_background.append(np.max(estimates.b[:,i].reshape(4800,1).dot(estimates.f[i,:].reshape(1,2000))))
+min_background = np.min(estimates.b, axis=0)*np.min(estimates.f, axis=1)
+max_background = np.max(estimates.b, axis=0)*np.max(estimates.f, axis=1)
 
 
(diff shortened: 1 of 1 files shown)


caiman/gui/gui_pyqtgraph_layout.py
```diff
@@ -46,13 +50,28 @@ def make_color_img(img, gain=255, out_type=np.uint8):
...
+max_mov_denoise = np.max(estimates.A.dot(estimates.C))
+min_background = []
+max_background = []
+background_num = -1
+neuron_selected = False
+nr_index = 0
+
+for i in range(0,estimates.f.shape[0]):
+    min_background.append(np.min(estimates.b[:,i].reshape(4800,1).dot(estimates.f[i,:].reshape(1,2000))))
+    max_background.append(np.max(estimates.b[:,i].reshape(4800,1).dot(estimates.f[i,:].reshape(1,2000))))
```
epnev: You can simply replace that with
```
min_background = np.min(estimates.b, axis=0)*np.min(estimates.f, axis=1)
max_background = np.max(estimates.b, axis=0)*np.max(estimates.f, axis=1)
```
more efficient plus does not depend on the dimensions.

caiman/gui/gui_pyqtgraph_layout.py
```diff
@@ -235,19 +348,146 @@ def mouseClickEvent(event):
...
+    bg_vline = pg.InfiniteLine(angle = 90, movable = True)
+    p2.addItem(bg_vline, ignoreBounds=True)
+    bg_vline.setValue(0)
+    bg_vline.sigPositionChanged.connect(show_background_update)
+    
+
+def show_background_update():
+    global bg_index, min_background, max_background, background_scaled
+    bg_index = int(bg_vline.value())
+    if bg_index > 0 and bg_index < 2001:
```
epnev: Instead of 2001 you should have something like `estimates.f.shape[-1]`.
Also is the indexing from 1 to T or is it pythonic from 0 to T-1?

caiman/gui/gui_pyqtgraph_layout.py
libbyli33: Thank you for the suggestion! I fixed and pushed it

caiman/gui/gui_pyqtgraph_layout.py
libbyli33: I agree. I believe it should be pythonic from 0 to T-1. Thx!

PR #580, opened 2019-07-16
Add a max_merge_area parameter for preventing the merging of components if the merged component's area is too large

jordan-g:
Currently, when performing CNMF, during the refinement stage overlapping components are merged if their correlation is above a certain threshold. This leads to situations like below, where two adjacent neurons whose activity is highly correlated can start off as two components but become merged into one component:

Screen Shot 2019-07-16 at 1 27 10 pm

In this case, splitting this single component back into two after-the-fact is a difficult problem. If the expected area of a neuron is known, this can be avoided by preventing components from being merged if the resulting merged component would have an area that's too large, even if the original components are highly correlated. This pull requests adds an optional `max_merge_area` parameter to do this. In the situation shown above, this allows us to have two components representing the two neurons, even when their activity is highly correlated:

Screen Shot 2019-07-16 at 1 28 26 pm

I think this is a useful addition as it allows more control over the automatic merging process during the refinement phase, especially when many tightly-packed neurons are active at the same time, which can otherwise lead to a single component spanning multiple neurons.

### DIFF
--- a/caiman/source_extraction/cnmf/cnmf.py
+++ b/caiman/source_extraction/cnmf/cnmf.py
@@ -94,7 +94,7 @@ class CNMF(object):
                  max_num_added=3, min_num_trial=2, thresh_CNN_noisy=0.5,
                  fr=30, decay_time=0.4, min_SNR=2.5, ssub_B=2, init_iter=2,
                  sniper_mode=False, use_peak_max=False, test_both=False,
-                 expected_comps=500, params=None):
+                 expected_comps=500, max_merge_area=None, params=None):
         """
         Constructor of the CNMF method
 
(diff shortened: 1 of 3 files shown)


epnev:
@jordan-g Thanks!

cawarwick:
Was max_merge_area intended to work with CNMFE? Because in my data it doesn't seem to prevent merging of dense/correlated components? The only caveat is I wasn't sure whether or not I needed to scale my 'area' into some other format beyond just pixels since the correlation maps I have are not in pixels, but in some arbitrary unit relative to the midline of X and Y. e.g. I measured my cells in pixels (e.g. a 14 x 14 pixel area) and used 200 max)merge_area.

image

PR #586, opened 2019-07-19
incomplete changes implementing 3d motion

as31:
rigid correction is mostly functional (through motion_correction.py, version in movies.py is incomplete). piecewise motion correction still needs to be debugged

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -206,6 +206,86 @@ class movie(ts.timeseries):
 
         return self, shifts, xcorrs, template
 
+    def motion_correct_3d(self,
+                       max_shift_z=5,
+                       max_shift_w=5,
+                       max_shift_h=5,
+                       num_frames_template=None,
+                       template=None,
+                       method='opencv',
+                       remove_blanks=False, interpolation='cubic'):
+        """
+        Extract shifts and motion corrected movie automatically,
+
+        for more control consider the functions extract_shifts and apply_shifts
+        Disclaimer, it might change the object itself.
+
+        Args:
+            max_shift,z,max_shift_w,max_shift_h: maximum pixel shifts allowed when 
+                    correcting in the axial, width, and height directions
+
+            template: if a good template for frame by frame correlation exists
+                      it can be passed. If None it is automatically computed
+
+            method: depends on what is installed 'opencv' or 'skimage'. 'skimage'
+                    is an order of magnitude slower
+
+            num_frames_template: if only a subset of the movies needs to be loaded
+                                 for efficiency/speed reasons
+
+
+        Returns:
+            self: motion corected movie, it might change the object itself
+
+            shifts : tuple, contains x, y, and z shifts and correlation with template
+
+            xcorrs: cross correlation of the movies with the template
+
+            template: the computed template
+        """
+
+        if template is None:  # if template is not provided it is created
+            if num_frames_template is None:
+                num_frames_template = old_div(
+                    10e7, (self.shape[1] * self.shape[2]))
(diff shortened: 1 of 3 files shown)


use_cases/NWB/demo_3d_motion_correction_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+                    "%(relativeCreated)12d [%(filename)s:%(funcName)20s():%(lineno)s]"\
+                    "[%(process)d] %(message)s",
+                    level=logging.WARNING)
+
+#%%
+def main():
+    pass  # For compatibility between running under Spyder and the CLI
+
+#%% Select file(s) to be processed (download if not present)
+    fnames = [os.path.join(caiman_datadir(), 'example_movies/sampled3dMovie.nwb')]
```
epnev: will this work with the file you created earlier?

caiman/base/movies.py
```diff
@@ -226,6 +306,26 @@ def bin_median(self, window=10):
...
+
+            window: int
+                number of frames in a bin
+
+        Returns:
+            img:
+                median image
+
+        """
+        d1, T, d2, d3 = np.shape(self)
```
epnev: We typically have the movie ordered as T, X, Y (,Z) with time in the first dimension. This is also what the CNMF expects. Is there any reason for this choice?

caiman/motion_correction.py
```diff
@@ -2098,6 +2243,272 @@ def tile_and_correct(img, template, strides, overlaps, max_shifts, newoverlaps=N
...
+            imgs, templates, [upsample_factor_fft] * num_tiles)]
+        shfts = [sshh[0] for sshh in shfts_et_all]
+        diffs_phase = [sshh[2] for sshh in shfts_et_all]
+        # create a vector field
+        shift_img_x = np.reshape(np.array(shfts)[:, 0], dim_grid)
+        shift_img_y = np.reshape(np.array(shfts)[:, 1], dim_grid)
+        shift_img_z = np.reshape(np.array(shfts)[:, 2], dim_grid)
+        diffs_phase_grid = np.reshape(np.array(diffs_phase), dim_grid)
+
+        #  shifts_opencv doesn't make sense here- replace with shifts_skimage
```
epnev: good point.

caiman/motion_correction.py
```diff
@@ -2098,6 +2243,272 @@ def tile_and_correct(img, template, strides, overlaps, max_shifts, newoverlaps=N
...
+        if shifts_opencv:
```
epnev: I would focus only `shifts_opencv = True`, at least for the beginning. It's simpler, faster (at least in 2d) and can easily handled motion fields upsampled to the original resolution. The only downside is that linear or cubic interpolation introduces a tiny amount of spatial smoothing that FFT interpolation does not. But I don't think this will play any role in practice.

epnev:
Thanks @as31 very impressive. I added a few comments.
Can I run the demo with the file you provided earlier?

caiman/base/movies.py
```diff
@@ -206,6 +206,86 @@ def motion_correct(self,
 
         return self, shifts, xcorrs, template
 
+    def motion_correct_3d(self,
```
epnev: The version in `movies.py` for the motion correction is deprecated. It's still used for rigid motion correction during online processing sometimes but I wouldn't worry about that.

caiman/base/movies.py
bendichter: NWB also expects T, X, Y (,Z)

PR #603, opened 2019-08-02
Hackathon2019

Hendriela:
Stuff that I did at the Flatiron. Includes 2 demo notebooks, a watershed ROI separation function, some documentation and parameter explanation for the wiki, and some post_analysis functions.

### DIFF
--- a/docs/source/Getting_Started.rst
+++ b/docs/source/Getting_Started.rst
@@ -64,7 +64,10 @@ crucial to be adapted to the specific dataset for proper analysis performance:
 
 -  ``fnames``: List of paths to the file(s) to be analysed. Memmap and hdf5 result files will be saved in the same directory.
 -  ``fr``: Imaging frame rate in frames per second.
--  ``decay_time``: Length of a typical transient in seconds. Default of ``0.4``. (half-decay time?)
+-  ``decay_time``: Length of a typical transient in seconds. ``decay_time`` is an approximation of the time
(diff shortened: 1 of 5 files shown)


docs/source/Getting_Started.rst
```diff
@@ -44,6 +44,70 @@ the following objects:
...
+-  ``quality``: Parameters for component evaluation (spatial correlation, SNR and CNN)
+-  ``spatial``: Parameters used in detection of spatial components
+-  ``temporal``: Parameters used in extraction of temporal components and deconvolution
+
+Of these parameters, most have a default value that usually does not have to be adjusted. However, some parameters are
+crucial to be adapted to the specific dataset for proper analysis performance:
+
+-  ``fnames``: List of paths to the file(s) to be analysed. Memmap and hdf5 result files will be saved in the same directory.
+-  ``fr``: Imaging frame rate in frames per second.
+-  ``decay_time``: Length of a typical transient in seconds. Default of ``0.4``. (half-decay time?)
```
Hendriela: Here I am not 100% sure how you define decay_time. It is the complete length of a transient, or the half-decay time (which is usually reported for fluorescence markers)? Following [the discovery paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3777791/), a transient of GCaMP6f for 1 AP lasts more than 1s, GCaMP6s even longer. If decay time actually means half-decay time, this should be addressed somewhere, otherwise the default of 0.4s might be too short for most markers.

epnev:
Thank you!

PR #611, opened 2019-08-05
3d motion changes 

### LINKED ISSUE #570: 3D Motion Correction
@bendichter Here are the basic steps needed to extend the motion correction algorithm to volumetric (3D) data. The main step is to write down a function that given a volume of data and a template, aligns the volume to the template.

To do that we need the following pieces:

- [ ] Break the volume and template into a sequence of overlapping sub-volumes and sub-templates. We have a function for this in [2D](https://github.com/flatironinstitute/CaImAn/blob/master/caiman/motion_correction.py#L1718) but should be easy to extend in 3D.
- [ ] Align each sub-volume to its corresponding sub-template. The [`register_translation_3d`](https://github.com/flatironinstitute/CaImAn/blob/master/caiman/motion_correction.py#L1220) method should be able to do this, although some testing is needed to ensure that it works accurately for sub-pixel registration.
- [ ] Align the inferred shifts from each sub-volume into a 3D grid. Upsample this grid to the original resolution of the 3D volume. Some care will be needed to there to ensure this is done correctly. For example, you can place each shift at the center of each subvolume and then interpolate/extrapolate between them. At the output, you should three volumes (one for each of the x-, y-, z-directions) that would locate how much in each direction has each pixel moved. In 2D we use the `cv2.resize` method which is not applicable to 3D. However, scikit-image has a method for [this](https://scikit-image.org/docs/dev/api/skimage.transform.html#resize).
- [ ] Apply the shifts to register the volume. Again in 2D we use the `cv2.remap` method which is not applicable to 3D, but scikit-image has the [`warp` method](https://scikit-image.org/docs/dev/api/skimage.transform.html#warp) that can do this job.

Let's start with this. Once this is done, we'll wrap around it the other methods we have for estimating template, correcting sequences of volumes, memory mapping etc.

as31:
Description
fix #570

demo_3d_motion_correction_pipeline.py has demo of 3d motion correction (rigid or pwrigid). 

Changed files:

Outstanding bugs: see TODO comments in motion_correction.py
motion_correction.py:
error after saving line 243 b0 = np.ceil 
parallelization needs to be readded

Fixes # (issue)

# Type of change

# Branching

# Has your PR been tested?


epnev:
Thanks @as31 
What's the relation between this PR and #586 ?
If this one is a superset, then please close #586 
If not, then commit your new changes on top of it.

bendichter:
@as31 

Running the demo I get the following error:

```python
  File "/Users/bendichter/dev/CaImAn/use_cases/NWB/demo_3d_motion_correction_pipeline.py", line 163, in <module>
    main()
  File "/Users/bendichter/dev/CaImAn/use_cases/NWB/demo_3d_motion_correction_pipeline.py", line 143, in main
    mc.motion_correct(save_movie=True)
  File "/Users/bendichter/dev/CaImAn/caiman/motion_correction.py", line 245, in motion_correct
    np.max(np.abs(self.z_shifts_els))))
TypeError: return arrays must be of ArrayType
```

as31:
This is the remaining bug that needs to be fixed, but the output memory
mapped file can be read with cm.load with the is3D flag for movie comparison

epnev:
@as31 Bug was caused because `np.maximum` cannot take more than two arguments. It should be fixed with https://github.com/flatironinstitute/CaImAn/pull/611/commits/c6796d6d6b120eaea4943e1975a3e7f20d7da0e0

Have you done any tests with respect to performance?

epnev:
@as31 I added another commit to fix a problem of not returning shifts in all 3 directions when performing rigid motion correction. The rigid registration looks fine. 

- For the pw-rigid the x- and y- directions are flipped. Can you fix that?
- We should test that pw-rigid does not fail miserably. A good way would be to create a dataset with only rigid motion and correct with pw-rigid. That should perform well and will ensure that the stitching and application of shifts is done correctly. Can you try that? Or does any of the datasets you created have only rigid motion?

as31:
The motion dataset with rigid in the name is rigid 3d motion only, and the
rigid only flag corrected that dataset well

epnev:
@as31 
So I run both rigid and pw-rigid in the file that has only rigid shifts. The pw-rigid didn't correct it properly. I took the pw-rigid shifts and computed their average for each frame in each direction and then computed the cross-correlation matrix with the rigid shifts. The result was
```
array([[ 0.93580632,  0.04914319,  0.22813308],
       [ 0.0103358 ,  0.95962393, -0.05561041],
       [-0.12107735,  0.05024465, -0.93954943]])
```
Even through the z-shifts are small this result indicates that somewhere a sign gets flipped. It's not clear where in the code at the moment.

as31:
Ah, dang. I'll look into it once I get back on the 21st

epnev:
@as31 I found what I believe was the bug. I'm planning to merge into dev but please take a look and do some tests once you're back. I expect an issue or two to pop up once people start to use it. Thanks for the outstanding work!

bendichter:
Alright! Thanks @epnev and @as31!

PR #612, opened 2019-08-05
numba + save initialization object

epnev:


caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -709,14 +751,22 @@ def fit_online(self, **kwargs):
...
+                Initialization can be passed as input: the initializer must be create by the same fit_online function
+
+            save_init_file: str
+                Path to the HDF5 file that will contain the initialization to be reused in the future
+
         Returns:
             self (results of caiman online)
         """
-        
+
```
epnev: @agiovann If you initialize with an `estimates` object the `init_batch` variable should be determined by the estimates, e.g. it should not be larger than the length of `C` inside estimates (it could be smaller, ideally they are the same).

demos/general/demo_OnACID_mesoscope.py
```diff
@@ -73,7 +71,7 @@ def main():
     # set up some additional supporting parameters needed for the algorithm
     # (these are default values but can change depending on dataset properties)
     init_batch = 200  # number of frames for initialization (presumably from the first file)
-    K = 2  # initial number of components
+    K = 50  # initial number of components
```
epnev: why the change in `K` and `gSig`?

demos/general/demo_OnACID_mesoscope.py
```diff
@@ -101,10 +99,18 @@ def main():
     opts = cnmf.params.CNMFParams(params_dict=params_dict)
 
 # %% fit online
-
+    init_file_save = fnames[0][:-4]+'_init.hdf5'
     cnm = cnmf.online_cnmf.OnACID(params=opts)
-    cnm.fit_online()
-
+    if os.path.exists(init_file_save):
```
epnev: including this feature as in the main demo add some complexity. Perhaps there is a better way.

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -18,6 +18,8 @@
 from builtins import range
 from builtins import str
 from builtins import zip
+#from numba import njit
```
epnev: I'm removing this since we're not using it yet (and it's not part of caiman's requirements).

pgunn:
I'm going to close this because I'm under the impression that a lot of these changes made their way into the codebase another way.

@caichangjia If this is not true, please reopen and we can chat further about it and figure out how to get it ready for merge.

PR #614, opened 2019-08-06
Some careful yapfing of some quiet files.

pgunn:
Maybe it's better to go over files in small sets so we can make sure yapf doesn't remove formatting that makes the code more legible and drop in enable/disables of it as needed.

### DIFF
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
 """ functions related to the creation and management of the cluster
 
 We put arrays on disk as raw bytes, extending along the first dimension.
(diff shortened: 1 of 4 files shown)


caiman/cluster.py
```diff
@@ -58,25 +62,24 @@ def extract_patch_coordinates(dims:Tuple, rf:Union[List,Tuple], stride:Union[Lis
...
-    iters = [list(range(rf[i], dims[i] - rf[i], 2 * rf[i] - stride[i])) + [dims[i] - rf[i]]
-             for i in range(len(dims))]
+    iters = [list(range(rf[i], dims[i] - rf[i], 2 * rf[i] - stride[i])) + [dims[i] - rf[i]] for i in range(len(dims))]
 
     coords = np.empty(list(map(len, iters)) + [len(dims)], dtype=np.object)
     for count_0, xx in enumerate(iters[0]):
         coords_x = np.arange(xx - rf[0], xx + rf[0] + 1)
         coords_x = coords_x[(coords_x >= 0) & (coords_x < dims[0])]
-        coords_x += border_pix*0 + np.maximum(sl_start[0], border_pix)
+        coords_x += border_pix * 0 + np.maximum(sl_start[0], border_pix)
```
pgunn: Whoa, wait. What's this multiplying by zero stuff?

PR #615, opened 2019-08-06
Fix 'shape mismatch in assignment' error in manual_merge function

jordan-g:
# Description

Using the `manual_merge` function leads to the following error in Python 3.7.3 due to `computedA` missing a  second dimension:

```
ValueError: shape mismatch in assignment
```

This is fixed by adding the missing dimension.

# Type of change


epnev:
thanks!

PR #618, opened 2019-08-07
Updates to fit_file

HagaiHargil:
# Description

A new `demo_pipe_as_func.py` file has been added which is completely identical to `demo_pipeline.py` in its function, but can also be called from anywhere else as a straight-forward function.

This is a preliminary proposal, if you like this general structure I might divide into more sub-functions (or perhaps a class...) and then add a new file (to this PR) which uses this function to parse multiple FOVs of the same animal. I will also link to the data we have for this example.

# Type of change


epnev:
thanks, will take a look when I get a chance.

HagaiHargil:
There's still an issue with the `.mmap` filename, and consequently, with the resulting `.hdf5` filename. It currently only includes the dimensions on the file currently being parsed, but if there are several files with the same dimensions in one folder, parsing them sequentially will keep overwriting the last `.mmap` and `.hdf5` files, preventing the use of the saved files for the preceding FOVs.

epnev:
@HagaiHargil We do have the `fit_file` function in the CNMF object that does everything up to the quality at one shot. Why not use that one, or extend it to cover the quality testing as well?

HagaiHargil:
I was unaware of `fit_file`, I will make a new PR that improves it. Do you want me to include a use case with our data?

HagaiHargil:
The PR for updating `fit_file` will be this one, since I wish to stay on the `dev` branch and not start a new fork.

I undid the previous changes and started "fresh".

I see that `fit_file` can accept more than one file (in the `fnames` list) but will only analyze the first one. Is this intended?

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -347,7 +347,9 @@ def fit_file(self, motion_correct=False, indices=None):
                 else:
                     b0 = np.ceil(np.max(np.abs(mc.shifts_rig))).astype(np.int)
                     self.estimates.shifts = mc.shifts_rig
-                b0 = 0 if self.params.get('motion', 'border_nan') is 'copy' else 0
+                # FIXME Huh?
```
HagaiHargil: This line specifically looks like a bug to me. Is it intended?

caiman/mmapping.py
```diff
@@ -73,7 +74,22 @@ def load_memmap(filename: str, mode: str = 'r') -> Tuple[Any, Tuple, int]:
             return (Yr, (d1, d2, d3), T)
     else:
         logging.error("Unknown extension for file " + str(filename))
-        raise Exception('Unknown file extension (should be .mmap)')
+        raise ValueError('Unknown file extension (should be .mmap)')
+
+    # Strip path components and use CAIMAN_DATA/example_movies
```
epnev: why is this code repeated here?

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -320,6 +320,8 @@ def fit_file(self, motion_correct=False, indices=[slice(None)]*2):
         Returns:
             cnmf object with the current estimates
         """
+        if indices is None:
```
epnev: Any reason for removing this from the function definition?

caiman/source_extraction/cnmf/cnmf.py
epnev: It specifies whether you want to exclude a number of pixels from the boundary. Right now it's always set to zero, but we might lift that in the future.

caiman/tests/test_memmapping.py
```diff
@@ -0,0 +1,19 @@
+from caiman import mmapping
```
epnev: I don't understand this test. Please explain.

caiman/source_extraction/cnmf/cnmf.py
HagaiHargil: Yes,

Mutable arguments in function definitions can result in very surprising behavior. See [this](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument?noredirect=1&lq=1) and [this](http://www.omahapython.org/IdiomaticPython.html#default-parameter-values), for example.

caiman/source_extraction/cnmf/cnmf.py
HagaiHargil: Is _now_ a good time to allow it...? :)

caiman/mmapping.py
HagaiHargil: That's an artifact from the `merge` operation I did, sorry about that.

caiman/tests/test_memmapping.py
HagaiHargil: I believe it should be clearer after my latest commit which fixes the merge artifact.

It's a simple unit test designed to make sure that `load_memmap` raises the proper error if given a non-`.mmap` file.

HagaiHargil:
I've added the component evaluation step to `fit_file`. It's placed under a flag since it seems to be messing up several tests and scripts that rely on the fact that `fit_file` doesn't do component evaluation.

I also added a couple more basic tests for the memory mapping phase, just because.

The next step was to create a use case for `fit_file` which runs several FOVs sequentially, but after some consideration it seems to be a bit too simple and specific, since all it will essentially be is a fancy wrapper around `fit_file`. In conclusion, I'm inclined to leave this PR as is, I believe it's valuable enough.

HagaiHargil:
I guess I need help in making sure that the two data files for the tests I added are including in the test builds.

caiman/source_extraction/cnmf/cnmf.py
pgunn: :+1:

caiman/source_extraction/cnmf/cnmf.py
epnev: While I like that attitude of doing things instead of talking about them, this is a bit more complicated. `b0` is equal to maximum shift observed across *all* directions and a non-zero value will exclude the same number of pixels from all directions which can be grossly suboptimal. Ideally we need to enable the option of making this direction specific. This can require changes in many places of the code to function properly. So let's leave it as a placeholder for the time being.

epnev:
> I guess I need help in making sure that the two data files for the tests I added are including in the test builds.

@pgunn might be able to help with the two data files being included in the test builds. In general I'm very hesitant with including data files in the distribution because I don't want to increase the size any further. However, these look pretty small so it should be fine. That being said, could we generate them on the fly and have them being loaded. That could also test the creation of the files.

HagaiHargil:
Are we positive we have write access to the `caiman_datadir()` location? If so, creating a setup function which generates these two files is very simple.

pgunn:
> Are we positive we have write access to the `caiman_datadir()` location? If so, creating a setup function which generates these two files is very simple.

Yes. That directory is generated by the user when they run `caimanmanager.py install` and it must be writable.

HagaiHargil:
The last semi-important change I made is to change the `base_name` of the memory-mapped files from `memmap_..` to `$ORIGINAL_FILENAME_memmap_...`. This was done because otherwise two recordings with the same FOV dimensions in the same folder will overwrite each other, rendering the save method and GUI kinda useless. I'm aware it's not completely related to this PR, so if you think that it should be moved to a different PR I'm fine with that.

epnev:
> The last semi-important change I made is to change the `base_name` of the memory-mapped files from `memmap_..` to `$ORIGINAL_FILENAME_memmap_...`. This was done because otherwise two recordings with the same FOV dimensions in the same folder will overwrite each other, rendering the save method and GUI kinda useless. I'm aware it's not completely related to this PR, so if you think that it should be moved to a different PR I'm fine with that.

That's a good point, thanks.

epnev:
Overall it looks fine. Just a final question: Are the memory mapped files created during the test or stored in /testdata? From the test, it doesn't look like they are created anywhere but they also don't appear in the list of changed files.

HagaiHargil:
> Overall it looks fine. Just a final question: Are the memory mapped files created during the test or stored in /testdata? From the test, it doesn't look like they are created anywhere but they also don't appear in the list of changed files.

I'm using automatic setup and tear-down function, as explained [here](https://nose.readthedocs.io/en/latest/writing_tests.html#test-functions), to create the file just before the test executes, and then delete it immediately after. During the running of the test they're located in `/testdata/`, but we're talking about a few milliseconds at most.

pgunn:
> Overall it looks fine. Just a final question: Are the memory mapped files created during the test or stored in /testdata? From the test, it doesn't look like they are created anywhere but they also don't appear in the list of changed files.

Longer-term we might want to think more about the where and the naming for memmap files, perhaps always putting them in a pid-named directory under /tmp by default? Collisions and poor naming are tied together. (this particular change is probably a step in the right direction)

epnev:
Perfect! Thanks @HagaiHargil 

@pgunn We are already doing something similar to what Hagai proposed in the other parts of the code. Otherwise we would have run into collisions multiple times in the past.

HagaiHargil:
The mem-mapped files are useful, since they're the motion-corrected version of the data. I regularly use them, and having them in `/tmp` defeats this purpose, I believe. They should live in the same folder as the raw data, which is what happens now, we just need to be a bit more careful regarding their naming.

PR #619, opened 2019-08-08
gui modification

caichangjia:
# Description

Add 'add new neuron' and 'merge neurons' options to the GUI.

Click 'add new neuron' to select possible place that existsneurons not found by CNMF, click 'add single' to confirm.

Click 'merge neurons' and select several neurons that you want to merge, click 'add single' to confirm change.

PR #621, opened 2019-08-09
yapf_2019_08_09

pgunn:
Ran yapf on files that haven't changed in the last 15 days or so.


caiman/tests/comparison_humans_online.py
```diff
@@ -101,26 +101,19 @@
...
-params_movie[1] = {
-    'folder_name': 'N.01.01/',
-    'ds_factor': 1,
-    'fr': 8,
-    'gnb': 1,
-    'decay_time': 1.4,
-    'gSig': [6, 6]
-}
+                                       # % neurofinder.01.01
+params_movie[1] = {'folder_name': 'N.01.01/', 'ds_factor': 1, 'fr': 8, 'gnb': 1, 'decay_time': 1.4, 'gSig': [6, 6]}
```
epnev: @pgunn 
kind of weird it chose to turn this into a single line but not the rest of the entries.

caiman/tests/test_temporal.py
```diff
@@ -10,12 +10,7 @@ def test_make_G_matrix():
...
     G = cnmf.temporal.make_G_matrix(T, g)
     G = G.todense()
-    true_G = np.matrix(
-        [[1., 0., 0., 0., 0., 0.],
-         [-1., 1., 0., 0., 0., 0.],
-         [-2., -1., 1., 0., 0., 0.],
-         [-3., -2., -1., 1., 0., 0.],
-         [0., -3., -2., -1., 1., 0.],
-         [0., 0., -3., -2., -1., 1.]])
+    true_G = np.matrix([[1., 0., 0., 0., 0., 0.], [-1., 1., 0., 0., 0., 0.], [-2., -1., 1., 0., 0., 0.],
```
epnev: older is easier to read here given it shows the matrix.

caiman/tests/comparison_humans_online.py
pgunn: I think the lack of a comment for this structure gave it some more flexibility. I should make how it does these consistent.

caiman/tests/test_temporal.py
pgunn: ok, will revert and prevent it from doing this again.

caiman/tests/comparison_humans_online.py
epnev: really minor points, no need to spend time on this.

epnev:
@pgunn I'll process first #612 and #619 and then come to this to avoid any potential conflicts. If you want to yapf a bit more do it directly here, although it might be prudent to wait until the code stabilizes reaches a steady state.

pgunn:
Whatever timing is easiest to do ; I can easily reconstruct the diff from scratch using just the tool and lists of things where we want to keep the existing formatting. If it makes it easier to merge I can redo this focused around files that haven't changed in a month rather than 10-ish days.

PR #640, opened 2019-09-09
Add a caimanmanager command to include coverage

pgunn:
This adds an (undocumented) subcommand to caimanmanager that runs nosetests with coverage enabled. It also adds coverage to environment.yml so source builds include it. I did not add it to the feedstock because there's no point in running this test outside of a source build.

Note as well that coverage imposes additional rules on python that restrict array resizing, causing caiman code to throw an error. We probably should look into this. The error looks like this:

```
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/caiman/tests/test_onacid.py", line 54, in test_onacid
    demo()
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/caiman/tests/test_onacid.py", line 46, in demo
    cnm.fit_online()
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/caiman/source_extraction/cnmf/online_cnmf.py", line 725, in fit_online
    self.initialize_online()
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/caiman/source_extraction/cnmf/online_cnmf.py", line 671, in initialize_online
    self._prepare_object(Yr, T1)
  File "/mnt/home/pgunn/miniconda3/envs/caiman/lib/python3.7/site-packages/caiman/source_extraction/cnmf/online_cnmf.py", line 180, in _prepare_object
    self.estimates.CY.resize([expected_comps + self.params.get('init', 'nb'), self.estimates.CY.shape[-1]])
ValueError: cannot resize an array that references or is referenced
by another array in this way.
Use the np.resize function or refcheck=False
```

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
 """ Suite of functions that help manage movie data
 
 Contains the movie class.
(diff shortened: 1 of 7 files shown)


pgunn:
(The mentioned error only shows up when caiman is run under coverage, not in other circumstnaces)

epnev:
@pgunn Can you fix that error? Could be as simple as storing `self.estimates.CY.shape[-1]` in a variable and then reading it from there.

pgunn:
Sure - I think we chatted briefly about this before you left and you said you'd want to look at this warning more closely to see if it was indicating a problem. I'll add refcheck=False to the call as that's the quickest way to make it happy.

pgunn:
@epnev Updated the diff - covtest is happy with the current changes.

PR #653, opened 2019-10-11
Dev fix gui

caichangjia:
# Description

Fix GUI rotation 90 degrees problem;
Adjust GUI such that it is compatible with .nwb file


epnev:
@caichangjia Great, thanks a lot!

PR #670, opened 2019-11-21
Move the static version into a single file, have other code consult that

pgunn:
This change makes the contents  /VERSION the single source of truth for static versions (on the granularity of a developer cutting a release, as opposed to commit hashes). It modifies setup.py to be aware of that versioning (meaning that for source installs, conda/pip understand the version of caiman correctly for the first time).

I have tested pip install and pip install -e on my workstation and both seem to do the right thing. I have also tested (in not great detail) the docs generation. Note that because the python package system is aware of versions with this diff, the best way to programmatically get at the version for the currently installed package is though:

```
import pkg_resources
...

pkg_resources.get_distribution('caiman').version
```
(same is true for any other installed package)


epnev:
looks good. Can you add a `caiman.__version__` method as well?

pgunn:
I added it as a string, as I'm guessing that's what you actually meant (and is what some other packages use). For some reason I thought setuptools did this for us if we gave it a good version string, and I was wrong.

PR #677, opened 2019-12-13
Use tifffile for .btf files

tbenst:
# Description

According to [OME-TIFF spec](https://docs.openmicroscopy.org/ome-model/6.0.1/ome-tiff/specification.html) `.ome.btf` is valid, and, in my experience, common. Since [tifffile](https://pypi.org/project/tifffile/) supports OME-TIFF, I've modified logic so CaImAn supports `.btf` files.

I've tested the motion correction pipeline on a very small `.btf` file and it seems to work properly. Unfortunately, most of my files are 70GB+ of `uint16` bigtiffs, so reading in as float32 means my computer ran out of memory. I plan on trying my hand at removing `asarray` & doing stream processing, but if someone at flatiron is already doing this please let me know so I don't duplicate effort! Thanks

# Type of change

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1510,6 +1510,7 @@ def load(file_name: Union[str, List[str]],
 
     if os.path.exists(file_name):
         _, extension = os.path.splitext(file_name)[:2]
+
         extension = extension.lower()
         if extension == '.mat':
             logging.warning('Loading a *.mat file. x- and y- dimensions ' +
@@ -1519,7 +1520,7 @@ def load(file_name: Union[str, List[str]],
             if mjv == 2:
                 extension = '.h5'
 
-        if extension == '.tif' or extension == '.tiff':        # load avi file
+        if extension in ['.tif', '.tiff', '.btf']:  # load avi file
             with tifffile.TiffFile(file_name) as tffl:
                 multi_page = True if tffl.series[0].shape[0] > 1 else False
                 if len(tffl.pages) == 1:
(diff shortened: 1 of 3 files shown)


epnev:
@tbenst Thanks for this. 
Is it possible to read only chunks of the `.btf` file without loading the entire dataset in memory?
E.g. by using something like `cm.load('file.btf', suindices=slice(100))`
If so, then caiman won't need to load the entire file during the standard patch processing pipeline.

tbenst:
@epnev great, that worked! So this pull-request is now tested and works fine for me.

I noticed a few jumps in the video--might this be happening on the border between slices, eg frame 99 to 100?

epnev:
@tbenst Sounds good, I'll be happy to incorporate this.
Since there are conflicts in several files can you sync your branch with the current version of our master or dev (branch) and then redo the PR? Dev is preferred but master is fine too.

epnev:
Regarding the jumps, if you want to load frames 0-99 and then 100-199 in two different steps you should do:
```
m1 = cm.load('file.btf', suindices=slice(100))
m2 =cm.load('file.btf', suindices=slice(100, 200))
```
Do you skip a frame if you do this?

pgunn:
@tbenst Hi, we haven't heard back from you on this pull request for awhile; it looks like you never got to eftychios's question. Can you resume the conversation to his satisfaction? If not we'll eventually want to close this without merging.

tbenst:
Hi @pgunn, rebased. The question on frame jump is still open but not related to this PR (it’s not a missing frame but rather seemed to me like the reference frame was the first subindex of each chunk).

epnev:
@pgunn good to merge.

PR #679, opened 2019-12-17
Added memory efficient function to remove baseline from movie using p…

agiovann:
Remove baseline using percentile method

# Description

Added a memory efficient method to remove baseline. This is based on a map method which only computes the median on temporal chunks of the movie. On my machine the dview=dview version does not work (it gets stuck). Suggesting to use dview=None. 

# Type of change

# Branching

- Pull requests with non breaking bug fixes should be made against the **master** branch.
- All other PRs should be made against the **dev** branch.

# Has your PR been tested?

--> **not tested** 

On small laptop and traveling


caiman/base/movies.py
```diff
@@ -38,6 +38,7 @@
 from typing import Any, Dict, List, Tuple, Union
 import warnings
 from zipfile import ZipFile
+from functools import partial
```
pgunn: let's keep our imports alphabetical

PR #680, opened 2019-12-17
support for efficiently removing baseline

agiovann:
# Description

+ Added support for efficiently removing baseline. 
+ Added  option to remove baseline in local correlation images
+ Updated demo_basic and demo_pipeline

Type
- [ X] New feature (non-breaking change which adds functionality)

# Has your PR been tested?
ran python caimanmanager.py test
**not tested on all demos** -> on small laptop


epnev:
@agiovann Can you do it against dev instead?

PR #687, opened 2020-01-03
Merge VolPy 

caichangjia:
# Description

Merge VolPy into CaImAn

1.  Add multiple new files for analyzing voltage imaging data
2.  Allow to load multiple .tif files
3.  Add argument masktasksperchild in setup_cluster
4.  Add compute_BL function which compute baseline function with less memory consumption


pgunn:
Hi,
Could you lowercase Volparams.py and spikePursuit.py?

caiman/source_extraction/volpy/mrcnn/config.py
```diff
@@ -0,0 +1,236 @@
+"""
+Mask R-CNN
+Base Configurations class.
+
+Copyright (c) 2017 Matterport, Inc.
+Licensed under the MIT License (see LICENSE for details)
```
pgunn: References a license file that's not included. This needs to be fixed.

pgunn:
I wish mrcnn had a conda package of its own so we wouldn't end up vendoring it with this diff. Particularly problematic as years later, if mrcnn has some bugfixes and other improvements, we'll still likely have this same version still stuck in our codebase that we don't maintain, pay attention to, or understand.

Is there any other way to do this?

epnev:
@caichangjia Add a reference to the preprint on the README file

caiman/source_extraction/volpy/mrcnn/config.py
caichangjia: So Pat, what should we do here?

Issue #700, opened 2020-01-27
cannot load GUI


vickerse1:
For better support, please use the template below to submit your issue. When your issue gets resolved please remember to close it.

Sometimes errors while running CNMF occur during parallel processing which prevents the log to provide a meaningful error message. Please reproduce your error with setting `dview=None`.

If you need to upgrade CaImAn follow the instructions given in the [documentation](https://caiman.readthedocs.io/en/master/Installation.html#upgrading).

- **Tell us a bit about your setup**: 
 1. Operating system (Linux/macOS/Windows): Windows 7
 
 2. Python version (3.x): 3.6.7
 
 3. Working environment (Python IDE/Jupyter Notebook/other): Anaconda prompt
 
 4. Which of the demo scripts you're using for your analysis (if applicable): demo_OnACID_mesoscope.py
 
 5. CaImAn version*: 1.7.1
 
 6. CaImAn installation process (`pip install .`/`pip install -e .`/conda): package based

*You can get the CaImAn version by creating a `params` object and then typing `params.data['caiman_version']`. If the field doesn't exist, type N/A and consider upgrading)

- **Describe the issue that you are experiencing**
Hello,
I successfully installed Caiman on Windows 7 with the package-based process, and was able to run the onACID mesoscope demo from a jupyter notebook. (it produced the output: MC_d1_440_d2_256_d3_1_order_F_frames_3000_.mmap)

However, when I attempt to load the GUI with the following command from your instructions:

ipython caiman/gui/gui_pyqtgraph_layout.py

I get this error:

ipython caiman/gui/gui_pyqtgraph_layout.py

I am in my caiman environment.

I have tried this from both the caiman directory:

ipython caiman/gui/gui_pyqtgraph_layout.py

and the caiman data directory:

C:\Users\scanimage\caiman_data

.....but I get the same error in both cases.

Am I in the wrong folder? Is there supposed to be a folder called "gui"? Do I need to update my version of Caiman to 1.8? Is 1.8 not available with package installation?

Thanks,

Evan

- **Copy error log below**

- **If you're not reporting an error, type your message below**

epnev:
@vickerse1 Can you put the actual error log? Right now you just copied the command.

@pgunn Can you take a look at this? It possible that the conda installation changes the location of the GUI

pgunn:
Hi,
Right now the GUI is experimental and the package version of CaImAn does not include an easy way to run it. Future versions of the conda package may make this easier.

--Pat

[...]

PR #701, opened 2020-01-29
make gui a command

pgunn:
This diff:
* Moves the current main GUI out of the library part of the code (nothing uses it as a library and it's a script). While doing so it gives it a better name
* Retires the ancient GUI that used to have that name to use_cases, which is more-or-less our junk drawer
* Adds a hashbang to the main GUI so people can launch it without saying python first
* Changes the install plumbing to make a "bin" subdirectory for caiman_data, which will be generally be a good place to put binaries of interest. The main GUI will live under ~/caiman_data/bin/
* Changes the GUI docs accordingly


pgunn:
This is meant to resolve issue #700

pgunn:
It also fixes a part of the docs that were not formatted right that I noticed while working on the diff

docs/source/GUI.rst
```diff
@@ -13,23 +13,23 @@ how use it, first load and run either \* demo_OnACID_mesoscope.py \*
...
 Each of these demos will save a results file. You can then start the
-visual interface by running the following command from the base caiman
-folder (make sure you are within your caiman environment):
+visual interface by visiting the caiman_data folder, under the bin folder,
+and issuing the following command:
 
 ::
 
-   ipython caiman/gui/gui_pyqtgraph_layout.py
+   python caiman_gui.py
```
epnev: same question as earlier: `python` or `ipython`?
@pgunn

pgunn: I tested it with regular python and it behaved fine. It doesn't seem to depend on ipython.

PR #703, opened 2020-01-31
volpy: keras is now an install optional

pgunn:
This diff:
* Notes that VolPy requires an optional install of Keras to function
* Removes keras from environment.yml
* Adds the volpy demo to things that the test commands in caimanmanager will never run

It should be followed up with a diff to the feedstock to take keras back out from the binary builds.


pgunn:
Adjusted the feedstock.

https://github.com/conda-forge/caiman-feedstock/commit/4d300974d16e058b4e34eebb234167450a19b4a3

PR #793, opened 2020-08-06
cnmf init_method fails for online_cnmf due to some estimates being None

draelos:
Moved estimates.g,bl,c1,lam outside init_method in initialize_online so method 'cnmf' can be used.

To replicate problem, running e.g. the demo_OnACID_mesocope.py fails if the init_method is changed from 'bare' to 'cnmf'. Fix assumes that all init_methods (bare, seeded, cnmf) use the same initialization of self.estimates.g, lam, c1, bl, and neurons_sn.

### DIFF
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -994,13 +994,7 @@ class OnACID(object):
                 (self.estimates.A, self.estimates.b, self.estimates.C, self.estimates.f,
                  self.estimates.YrA) = tmp
             self.estimates.S = np.zeros_like(self.estimates.C)
-            nr = self.estimates.C.shape[0]
-            self.estimates.g = np.array([-np.poly([0.9] * max(self.params.get('preprocess', 'p'), 1))[1:]
-                               for gg in np.ones(nr)])
-            self.estimates.bl = np.zeros(nr)
-            self.estimates.c1 = np.zeros(nr)
-            self.estimates.neurons_sn = np.std(self.estimates.YrA, axis=-1)
-            self.estimates.lam = np.zeros(nr)
+            
         elif self.params.get('online', 'init_method') == 'cnmf':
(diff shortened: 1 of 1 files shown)


pgunn:
The code looks good (and reducing duplicated code is a win). @agiovann Does this look like a good change from a larger view?

j-friedrich:
It does not fail, as long as you make sure to set `only_init` to `False` (or use patches). The idea behind `only_init` was to merely run the initialization on patches and do the update_temporal, update_spatial, ... iterations when the patches are combined. We could consider setting `only_init` to `False` automatically if no patches are used (`rf==None`), to ensure the entire pipeline is run. If one really wants to perform only  the initialization, one can just use `cnm.initalize(Y)`.

The 'cnmf' initialization estimates the time constants for each neuron detected in the initial batch individually based on the data, and uses the mean time constant for all neurons detected during the online phase. 'seeded' and 'bare' use fixed hard coded values, at the moment 0.9 for p=1. These should probably better be determined as a function of the parameters `fr` and `decay_time`

draelos:
Thanks for the explanation! The 'cnmf' initialization was suggested for data with lots of active neurons, which I have, so I certainly want to use it. I'm currently trying to run initialize_online() with a small set of frames (e.g. 50) and then run fully online (e.g. starting with frame 51). Setting only_init to False does work, but does your comment about just using cnm.initialize(Y) mean I could be using that, and then seamlessly run with fit_online()? Looks like if rf is None and estimates.A is None then initialize(Y) gets run automatically regardless of the only_init parameter, and doesn't set estimates.g, etc. 

If cnmf as an init_method requires only_init to be False (if rf=None) then I agree perhaps that should be enforced instead of failing later at the creation of OASISinstances.

pgunn:
@draelos @j-friedrich Where did things land on this in your conversation?

pgunn:
@j-friedrich @draelos Any news on this? We should figure out whether to merge this or not.

pgunn:
Closing for inactivity

PR #805, opened 2020-08-31
Support loading videos from MKV containers

ximion:
# Description

In our lab we make extensive use of the more modern Matroska container (MKV) for videos, mostly containing FFV1 or VP9-encoded material. The new Miniscope v4 DAQ software as well as the PoMiDAQ Miniscope DAQ software are also either using MKV as default or supporting it in combination with FFV1 as an option.

Fortunately, making CaImAn support loading video data from MKV files natively is a very trivial change, so with this patch it can load data generated in the MKV format directly without any conversion.
All Linux distributions understand MKV natively (via GStreamer or FFmpeg) and Windows supports this format out of the box since Windows 10 as well.

# Type of change

# Has your PR been tested?

Yes, it doesn't break anything and works fine for our files (CaImAn loads and analyzes them)

Thanks for considering!

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1562,7 +1562,7 @@ def load(file_name: Union[str, List[str]],
 
                 input_arr = np.squeeze(input_arr)
 
-        elif extension == '.avi':      # load avi file
+        elif extension in ('.avi', '.mkv'):      # load video file
             cap = cv2.VideoCapture(file_name)
 
             try:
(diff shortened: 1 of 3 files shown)


pgunn:
I'm a little wary of those dependencies (they might not be there on headless/cluster installs), and am unsure if OSX handles it as seamlessly, but this is a small change and should not cause any harm.

ximion:
> I'm a little wary of those dependencies (they might not be there on headless/cluster installs), and am unsure if OSX handles it as seamlessly [...]

Our HPC cluster running CentOS 6 supports this easily (but that also pulls in a pretty recent ffmpeg via conda), and GStreamer and FFmpeg pretty much support MKV since forever - FFmpeg supports it since at least 2007, and OpenCV supports it for at least 7 years explicitly, which was when MKV was mentioned in the initial tutorial. MacOS for some reason only supports MKV's WebM derivative in Safari, apparently, but people will have FFmpeg on macOS anyway. And even if the container wasn't supported, OpenCV will give a pretty clear error message.
So I would be immensely surprised if this would cause any issue or even confusion in future :-)

Thank you for merging this change!

PR #806, opened 2020-08-31
Fix sentence in demo_motion_correction.ipynb

ckerem:
- [ x] Changes in documentation or new examples for demos and/or use cases.


pgunn:
@ckerem Thanks for spotting this!

Issue #711, opened 2020-02-13
Tracking issue to make python 3.8 builds on conda-forge


pgunn:
This is a tracking issue to
1) Test caiman in python 3.8 environments managed by conda, and when it passes those tests
2) Make any code changes to Caiman needed for 3.8 compatibility
3) Change the caiman feedstock to include a python 3.8 build
4) Deprecate python 3.6 (announcement)
5) Some time after step 4, change the caiman feedstock to stop building python 3.6 packages

pgunn:
Right now the environment is not buildable, either with the default pins or with unpinning everything except for python=3.8. Current task is to find/fix this.

pgunn:
Chatted with Anaconda devs - apparently there's no build of Tensorflow for Python 3.8 yet (anaconda or not). We'll need to wait a bit.

pgunn:
Still no build for 3.8

pgunn:
Recently a build of Tensorflow for Python 3.8 was packaged. This is no longer blocked (at least by this); I'll be trying a 3.8 build soon.

pgunn:
I was just able to manually build a 3.8 environment of caiman with small changes to the environment file, including unpinning everything. Will be testing it and if things look good we'll be able to lift the pins and add 3.8 builds to the conda-forge package of caiman.

pgunn:
Passes nosetests

pgunn:
Passes demotests. Upcoming PR to main codebase to enable all of this also cleans up some bad code that uses "is" where it should use "==".

pgunn:
We've been waiting for py38 builds of tensorflow, which arrived early on Linux and very recently on Windows. Still no OSX builds though, so we're going to need to wait.

pgunn:
For the curious, this is what I'm keeping an eye on:

https://anaconda.org/anaconda/tensorflow/files

As soon as we see osx py38 builds in there, we'll be good to get moving on this again.

pgunn:
We wrapped this up recently.

PR #810, opened 2020-09-10
Python 3.8

pgunn:
This makes necessary adjustments to the codebase to make py38 the normal path for people who build an environment with everything-but-caiman using conda, and who then use pip to compile conda. This is the "dev" workflow.

It passes nosetests and demotests.

This implements half of #711 (the other half will be changing the feedstock).


pgunn:
Unfortunately, this is premature; there are no tensorflow builds for py38 yet on MacOS and Windows.

pgunn:
There are now py38 builds for Windows; still no Mac builds so we're still waiting.

pgunn:
Depends on https://github.com/ContinuumIO/anaconda-issues/issues/11697

PR #934, opened 2021-11-17
Updates load in movies.py

JakeHeffley:
Updates the load function so it uses memory more efficiently when loading a large tiff file stored as a single pagefile.

# Description

When loading a large tiff stored as a single pagefile movies.load() will load the entire tiff file into each active processor. This tends to use up all the available memory. 

Fixes # (issue)

In tffl.asarray() set out = 'memmap'    This uses memory more efficiently but is still slow. 

- [* ] Bug fix (non-breaking change which fixes an issue)

# Has your PR been tested?
Yes, test and demotest

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1544,7 +1544,7 @@ def load(file_name: Union[str, List[str]],
                                                                       np.arange(shape[1]))
                                                          ).reshape((len(ts),) + shape[1:])
                         else:
-                            input_arr = tffl.asarray()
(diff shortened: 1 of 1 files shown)


pgunn:
I believe this is the right thing to do (it changes behaviour slightly in that the default of None means writes to the data won't be pushed down or create sync concerns, but I believe that won't be an issue here).

I may at some point adjust this to make the new flag optional as a performance tuning thing.

(if we find any limits on what can be done with a memmapped array, I may need to revert, but I don't expect that).

Thanks for the patch

PR #944, opened 2021-12-17
update to keep data as float32

JakeHeffley:
# Description
Fixes an issue which arose recently where dtype becomes float64 for the estimates (A and C). This causes an inefficient use of memory. e.g. when reconstructing the denoised movie.

# Type of change
Specified dtype=np.float32 as an argument for several numpy functions. 

- [ *] Bug fix (non-breaking change which fixes an issue)

# Has your PR been tested?
Yes, test and demotest with caimanmanager


pgunn:
I will need to study this to make sure any potential loss in accuracy isn't too severe. Thank you for the PR.

pgunn:
I spoke with Andrea about this diff and he's concerned that it will have unintended consequences; at least in the near future I don't expect we'll be able to merge this. If you'd like you can chat with him on our slack, but for now I'm closing the issue.

PR #950, opened 2022-02-17
Allow custom basename when saving mcorr memmaps

kushalkolar:
# Description

Currently, if the same input file is put through motion correction with different parameter variants the resulting motion corrected memmap will always have the same filename. I could not find a way to set the filename of mcorrr memmaps that are saved when setting `save_movie=True` in `MotionCorrect.motion_correct()`.

I added a kwarg called `base_name_prefix` to `MotionCorrect.motion_correct()` which allows a user to append a prefix to the basename that is used when saving memmaps. By default it is set to `None`, but if a user supplies a `str` it will set that as a prefix to the `base_name` when saving the memmap.

Let me know if there's a better way to do this!

# Type of change

# Branching
targets `dev`

# Has your PR been tested?

Tested that currently functionality isn't broken by running `test_demos.sh`, no failures. Also tested the new functionality and it does work, prefixes are set to saved memmaps.


kushalkolar:
closing because this has been replaced by #978 which has already been merged.

PR #965, opened 2022-04-05
volpy maskrcnn compatible tensorflow 2.4.1

caichangjia:
# Description

Mask R-CNN in VolPy previously only worked on tensorflow 1 and needed users to install keras. Now it worked on tensorflow2.4.1.

Fixes # (887, 963)

# Type of change

### DIFF
--- a/caiman/source_extraction/volpy/mrcnn/model.py
+++ b/caiman/source_extraction/volpy/mrcnn/model.py
@@ -8,37 +8,33 @@ Written by Waleed Abdulla
 """
 
 import os
-import random
 import datetime
 import re
 import math
-import logging
 from collections import OrderedDict
 import multiprocessing
 import numpy as np
 import tensorflow as tf
-#import tensorflow.keras as keras
-#import tensorflow.keras.backend as K
-#import tensorflow.keras.layers as KL
-#import tensorflow.python.keras.engine as KE
-#import tensorflow.keras.models as KM
-import keras
-import keras.backend as K
-import keras.layers as KL
-import keras.engine as KE
-import keras.models as KM
+import tensorflow.keras as keras
+import tensorflow.keras.backend as K
+import tensorflow.keras.layers as KL
+import tensorflow.keras.layers as KE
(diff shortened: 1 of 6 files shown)


caiman/source_extraction/volpy/mrcnn/model.py
```diff
@@ -2378,7 +2368,7 @@ def train(self, train_dataset, val_dataset, learning_rate, epochs, layers,
             validation_steps=self.config.VALIDATION_STEPS,
             max_queue_size=100,
             workers=workers,
-            use_multiprocessing=True,
+            use_multiprocessing=workers > 1,
```
pgunn, line 2371: I see above you mention that Keras fails on Windows when using multiprocessing; is this still true of the keras layer in tensorflow? If not, the comment above (line 2363) should be removed. If not, is this code nonusable on windows with this change?

pgunn (commented):
Generally looks solid, and it's great news that we won't be stuck on TF1.x.

I left a few comments.

caiman/source_extraction/volpy/mrcnn/model.py
```diff
@@ -8,38 +8,34 @@
...
+import tensorflow.keras.layers as KL
+import tensorflow.keras.layers as KE
+import tensorflow.keras.utils as KU
+from tensorflow.python.eager import context
+import tensorflow.keras.models as KM
 
 from ..mrcnn import utils
 
-# Requires TensorFlow 1.3+ and Keras 2.0.8+.
+# Requires TensorFlow 2.0+
```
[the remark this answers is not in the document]
caichangjia: You mean update tensorflow to 2.0+ in the environment.yml file? Yes, we can do this. 
We are using tensorflow 2.4.1 in the current version of CaImAn anyway.

caiman/source_extraction/volpy/mrcnn/model.py
caichangjia: I actually only tested the Mask R-CNN in VolPy on the linux system.

PR #973, opened 2022-05-07
3D Piecewise Rigid Motion Correction Respecting value of border_nan

GBarnum:
Added support for other values of border_nan in tile_and_correct_3d() and MotionCorrect.apply_shifts_movie() when using piecewise rigid motion correction.

# Description

Previously, when using piecewise rigid motion correction on 3D data, the border was always filled with `0.0` regardless of the value of `border_nan` set. This change results in `np.nan` in the border when `border_nan = True`, copying of the border values when `border_nan = 'copy'`, and filling with the minimum of the image when `border_nan = 'min'`. The previous behavior is used when `border_nan = False`.

# Type of change

# Has your PR been tested?

Both

```python caimanmanager.py test```

and

```python caimanmanager.py demotest```

run and report success.


pgunn:
Sorry for the delay; we're looking into this now.

PR #1003, opened 2022-08-15
Fix cropping method and add ability to save compressed .avi files with the FFV1 codec

emelon8:
# Description

This pull request fixes the cropping method in the movie class, which gives a ValueError when run (e.g., "ValueError: could not broadcast input array from shape (1000,340,348) into shape (1000,390,388)"). The changes make it so the cropped movie is returned and not cropped inline.

A second edit also allows timeseries.save to save .avi files using the FFV1 codec, which a lossless codec, resulting in a much smaller file size.

# Type of change

# Has your PR been tested?

It had one error running python caimanmanager.py test: the parallel test. I'm not sure what the issue is, since my edits are minor and shouldn't affect the parallel processing. I've recently been getting parallel errors with my computer (prior to any edits of the code), though, so that's likely the issue.

The demotest ran successfully.

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -549,19 +549,15 @@ class movie(ts.timeseries):
 
     def crop(self, crop_top=0, crop_bottom=0, crop_left=0, crop_right=0, crop_begin=0, crop_end=0) -> None:
         """
-        Crop movie
+        Crop movie (inline)
 
         Args:
             crop_top/crop_bottom/crop_left,crop_right: (undocumented)
 
             crop_begin/crop_end: (undocumented)
-            
-        Return:
-            cropped movie
         """
         t, h, w = self.shape
-        cropped_movie = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
-        return cropped_movie
+        self[:, :, :] = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
     
     def removeBL(self, windowSize:int=100, quantilMin:int=8, in_place:bool=False, returnBL:bool=False):                   
         """
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -549,15 +549,19 @@ class movie(ts.timeseries):
 
     def crop(self, crop_top=0, crop_bottom=0, crop_left=0, crop_right=0, crop_begin=0, crop_end=0) -> None:
         """
-        Crop movie (inline)
+        Crop movie
 
         Args:
             crop_top/crop_bottom/crop_left,crop_right: (undocumented)
 
             crop_begin/crop_end: (undocumented)
+            
+        Return:
+            cropped movie
         """
         t, h, w = self.shape
-        self[:, :, :] = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
+        cropped_movie = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
+        return cropped_movie
     
     def removeBL(self, windowSize:int=100, quantilMin:int=8, in_place:bool=False, returnBL:bool=False):                   
         """
--- a/caiman/base/timeseries.py
+++ b/caiman/base/timeseries.py
@@ -160,6 +160,10 @@ class timeseries(np.ndarray):
             q_max, q_min: float in [0, 100]
                 percentile for maximum/minimum clipping value if saving as avi
                 (If set to None, no automatic scaling to the dynamic range [0, 255] is performed)
+                
+            compress: int
+                if saving as .tif, specifies the compression level
+                if saving as .avi or .mkv, compress=0 uses the IYUV codec, otherwise the FFV1 codec is used
 
         Raises:
             Exception 'Extension Unknown'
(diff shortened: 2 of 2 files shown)


caiman/base/movies.py
```diff
@@ -549,15 +549,19 @@ def linf(x, a, b):
...
 
             crop_begin/crop_end: (undocumented)
+            
+        Return:
+            cropped movie
         """
         t, h, w = self.shape
-        self[:, :, :] = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
+        cropped_movie = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
+        return cropped_movie
```
pgunn, line 564: I like this change; it's something I've wanted to fix broadly across the codebase (there are other methods that modify self, sometimes also returning it for some reason), but it takes some work to make sure that nothing depends on the current (bad) behaviour. I'll try to figure that out for this before accepting the PR.

emelon8: Excellent, thanks! I should also give credit to @IsaacJLambert, the student who suggested that change. :)

pgunn: Right now this change won't work - the calls to the crop method that I can find in the codebase depend on the inline behaviour, ignoring the return value.

This leaves us with two good options to make the code here cleaner:
1) We adjust those calls to do assignment when they do a crop, which works but is most useful if we expect people will want to keep both cropped and uncropped objects available in their code (might be a little unlikely)
2) We adjust this function not to return anything, becoming purely an inline shift

The second is very easy; I don't see any of our code that ever cares about the return value of the crop method, so we could just remove the return value

Thinking about it, right now I think I prefer the second, although I'm open to arguments as to why the first is better I'm happy to listen.

Otherwise, if you want to remove this part of the diff and optionally remove the return value from crop that'd be great.

pgunn: (by "not work", I mean this diff turns the crop function into a no-op by having existing code discard the cropped version of the data)

emelon8: I agree that an inline crop would be ideal. Maybe I'm not understanding something, but the existing inline crop method hasn't been working for me. It's giving a ValueError when I run it, since I'm trying to put a 1000 x 340 x 348 array into the existing 1000 x 390 x 388 array. Is there an alternative way to do the cropping inline?

pgunn: Oh, I think mixing the design of your code and the existing code should fix that. Having the assignment be to self rather than self[:,:,:] should solve the mismatch.

emelon8: I had the same idea, but unfortunately, I don't think that will work either. When the assignment is to `self` (rather than `self[:,:,:]`), it is cropped as expected (I can see it while running in debug mode); but when the method finishes, the movie object is unchanged. For instance,
```Python
print(mymovie.shape)
mymovie.crop(crop_top=10, crop_bottom=40, crop_left=10, crop_right=30)
print(mymovie.shape)
```
returns
```Python
(1000, 390, 388)
(1000, 390, 388)
```
I think it's the same issue described here: https://stackoverflow.com/questions/64668170/why-changing-object-itself-through-class-object-doesnt-work. If there were specific attributes we could change, then I think it would work.

pgunn: That's right; I didn't think this through correctly.

pgunn: One way to do it would be to do something like this:
```
self.resize( (new,dims,here), refcheck=False)
self[:,:,:] = new_data
```

emelon8: This doesn't work right now, either. Turns out the np.ndarray.resize method has been overloaded with a resize method in CaImAn's movie class that doesn't do the same thing. Do you know of any other way to resize `self` inline?

pgunn: (I'll review the few calls in the codebase to this method and see if I can figure out what they think they're getting from using it, because as I see it right now the original method is either a no-op or will fail)

emelon8: I'm still figuring CaImAn out, so there may be a better way to do it, but I would use this method to crop our movies before we start the other steps in the pipeline (motion correction, CNMF-E): I would load the original movie, crop it, save it back to disk, and then run the rest of the pipeline on the cropped movie.

pgunn:
Ah, for that I'd suggest preprocessing it in some other package, like ImageJ, before feeding it to caiman.

I'll get back to you on the patch later today; I suspect I'll ask you to chop the first part out (and I'll followup later on the topic), but I'd like to read the code that uses the function first to see if I can figure out what it was trying to do and if this ever worked.

caiman/base/movies.py
pgunn: Right now I'm pretty sure this code never worked and was never used; I'm inclined to remove it rather than try to fix it. I suggest doing this kind of cropping before loading data into caiman, as doing it in a notebook or other code risks other parts of the code learning things about the broader dimensions and after a crop not learning that the dimensions changed.

If you don't mind revising the PR to just cover the movie file format improvements, that'd be ideal.

(thanks for calling my attention to this though; removing old broken code is a good thing)

emelon8: OK, I've reverted the PR so it just includes the first commit with the movie file formats. I haven't removed the cropping method--it's just as it was before.

For what it's worth, we've been using a copy of the cropping method in our own code to crop the movie object and return the cropped movie. So far, we haven't seen any malfunction in running motion correction or CNMF-E on the saved cropped movie.

pgunn:
If it's useful, I'll solve this a slightly different way; I'll remove all existing calls to the crop method, rename it, and make it available as a convenience function.

PR #1004, opened 2022-08-22
Make crop() function useful (with new name)

pgunn:
See #1003

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -187,7 +187,7 @@ class movie(ts.timeseries):
         self = self.apply_shifts(shifts, interpolation=interpolation, method=method)
 
         if remove_blanks:
-            raise Exception("The remove_blanks parameter was never functional and should not be used")
+            raise Exception("motion_correct(): The remove_blanks parameter was never functional and should not be used")
 
         return self, shifts, xcorrs, template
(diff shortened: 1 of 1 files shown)


caiman/base/movies.py
```diff
@@ -535,7 +535,7 @@ def return_cropped(self, crop_top=0, crop_bottom=0, crop_left=0, crop_right=0, c
             crop_begin/crop_end: (undocumented)
         """
         t, h, w = self.shape
-        ret = np.resize(self, ( t - crop_end - crop_begin, h - crop_bottom - crop_top, w - crop_right - crop_left))
+        ret = np.zeros(( t - crop_end - crop_begin, h - crop_bottom - crop_top, w - crop_right - crop_left))
         ret[:,:,:] = self[crop_begin:t - crop_end, crop_top:h - crop_bottom, crop_left:w - crop_right]
```
emelon8, line 540: Is there a reason to cast the movie object into a ndarray vs. keeping it as a movie since it is not cropping inline (and thus shouldn't have any effects downstream)?

pgunn: It doesn't much matter (and it should be trivial to pull it back into being a movie if desired).

PR #1014, opened 2022-10-13
mmapping paths: rework parsing of filenames

pgunn:
This changes the naming semantics (slightly) and parsing of memmap files to make them less fragile to oddly-named input files and to make them easier to read/maintain.

It has the potential to be disruptive.


kushalkolar:
I was wondering why memmap path sanity check tests were failing in `mesmerize-core` tests and then came across this, I noticed that it's been incorporated into the dev branch so I guess it's going to be merged into master eventually? Shouldn't be an issue for users downstream, I can update our tests once its merged.

Basically seems like trailing underscores are removed?

current master branch:
`memmap__d1_128_d2_128_d3_1_order_C_frames_1000_.mmap`

dev:
`memmap_d1_128_d2_128_d3_1_order_C_frames_1000.mmap`

pgunn:
@kushalkolar Yup. If there are any other filename-related cleanups you'd like to suggest we mix in here, maybe also utility functions you might like for mesmerise that help generate/parse them, I would be happy to take a look. Sloppy use of filenames felt like a historic weakness of the codebase that I'm working on cleaning up as I hope to eventually do more ambitious changes to how intermediate files are named/placed.

kushalkolar:
Our sanity checks are just to confirm filepaths referenced in the dataframe end up in the right place in the batch item's directory and that they are resolvable, so as long as we know the new system it'll be fine. Only issue I could see is if users have used the memmap file names in their own parsing code. I guess memmaps don't support anything like fileheaders to store this info instead.

pgunn:
Maybe we could make an API-friendly  version of most of the entry points where one passes in a path config in the params object, and one gets back a list of filenames in a retrurn structure?

kushalkolar:
yea, this is kinda why I created mesmerize because it kinda does this :)

kushalkolar:
doing it at the caiman-level would make things better and easier though

Issue #1021, opened 2022-12-10
which `nose` is required to run `caimanmanager.py test`


kushalkolar:
This is missing from the contributing docs, https://github.com/flatironinstitute/CaImAn/blob/master/CONTRIBUTING.md 

running `python ./caimanmanager.py test` in the root dir with a symlink install, i.e. `pip install -e .`
with only `nose` installed I get

```python3
Traceback (most recent call last):
  File "/home/kushalk/python-venvs/mescore/bin/nosetests", line 8, in <module>
    sys.exit(run_exit())
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/core.py", line 118, in __init__
    unittest.TestProgram.__init__(
  File "/usr/local/lib/python3.10/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/core.py", line 179, in parseArgs
    self.createTests()
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/core.py", line 193, in createTests
    self.test = self.testLoader.loadTestsFromNames(self.testNames)
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/loader.py", line 481, in loadTestsFromNames
    return unittest.TestLoader.loadTestsFromNames(self, names, module)
  File "/usr/local/lib/python3.10/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/local/lib/python3.10/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/loader.py", line 431, in loadTestsFromName
    return self.loadTestsFromModule(
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/loader.py", line 354, in loadTestsFromModule
    tests.extend(self.loadTestsFromDir(module_path))
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/loader.py", line 191, in loadTestsFromDir
    yield self.loadTestsFromName(
  File "/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/nose/loader.py", line 431, in loadTestsFromName
[...]

PR #1022, opened 2022-12-10
sklearn deprecation, remove `normalize` from LassoLars

kushalkolar:
# Description

minor change, the latest sklearn v1.2 that just released deprecated the `normalize` kwarg for `LassoLars`, it just produces a massive stream of warnings during CNMF (they can be disabled but it'll be removed by v1.4 so might as well fix it?)

```python3
/home/kushalk/python-venvs/mescore/lib/python3.10/site-packages/sklearn/linear_model/_base.py:117: FutureWarning: 'normalize' was deprecated in version 1.2 and will be removed in 1.4. Please leave the normalize parameter to its default value to silence this warning. The default behavior of this estimator is to not do any normalization. If normalization is needed please use sklearn.preprocessing.StandardScaler instead.
  warnings.warn(
```

not exactly sure which tests should be used, I ran `nose2` in the root dir with both `nose` and `nose2` installed and the tests pass, see #1021

### DIFF
--- a/caiman/source_extraction/cnmf/spatial.py
+++ b/caiman/source_extraction/cnmf/spatial.py
@@ -410,7 +410,7 @@ def regression_ipyparallel(pars):
                 model = make_pipeline(
                     StandardScaler(with_mean=False),
                     linear_model.LassoLars(alpha=lambda_lasso, positive=True,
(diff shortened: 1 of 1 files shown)


pgunn:
Thanks!

PR #992, opened 2022-06-30
Fix cnmf utils get file size for tiff files created in append mode

kushalkolar:
# Description

Tiff files created in append mode raise an IndexError when saving the memmaped motion correction movie. This is due to `cnmf.utilities.get_file_size()` returning the wrong dims.

Example: 

```python
# tiff file written in append mode
cm.source_extraction.cnmf.utilities.get_file_size("/home/kushal/tiff_mcorr_issues/ap1_1_pt_5a.tif")
>>> ((216,), 270)

# non-append mode but still multi-page tiff
cm.source_extraction.cnmf.utilities.get_file_size("/home/kushal/tiff_mcorr_issues/ap1_1_pt_5a_single.tif")
>>> ((270, 216), 27090)
```

Tiff files created in append mode are created like this:

```python
tif_writer = tifffile.TiffWriter(output_path, bigtiff=True, append=True)

for frame_ix in range(n_frames):
    # do stuff to generate my output frame
    tif_writer.write(frame)
```

Tests passing on my end.

Stack trace without the fix:
```python
IndexError                                Traceback (most recent call last)
Input In [9], in <cell line: 2>()
      1 # correct for rigid motion correction and save the file (in memory mapped form)
----> 2 mc.motion_correct(save_movie=True)

File ~/repos/CaImAn/caiman/motion_correction.py:267, in MotionCorrect.motion_correct(self, template, save_movie)
    264         b0 = np.ceil(np.maximum(np.max(np.abs(self.x_shifts_els)),
    265                             np.max(np.abs(self.y_shifts_els))))
    266 else:
--> 267     self.motion_correct_rigid(template=template, save_movie=save_movie)
    268     b0 = np.ceil(np.max(np.abs(self.shifts_rig)))
    269 self.border_to_0 = b0.astype(np.int)

File ~/repos/CaImAn/caiman/motion_correction.py:301, in MotionCorrect.motion_correct_rigid(self, template, save_movie)
    298 self.shifts_rig:List = []
    300 for fname_cur in self.fname:
--> 301     _fname_tot_rig, _total_template_rig, _templates_rig, _shifts_rig = motion_correct_batch_rigid(
    302         fname_cur,
    303         self.max_shifts,
    304         dview=self.dview,
    305         splits=self.splits_rig,
    306         num_splits_to_process=self.num_splits_to_process_rig,
    307         num_iter=self.niter_rig,
    308         template=self.total_template_rig,
    309         shifts_opencv=self.shifts_opencv,
[...]

PR #1041, opened 2023-02-08
Open multiseries tiff files and nd2 files

RuiOSilva:
# Description

Add functionality to work with multiseries tif files and nd2 (from Nikon softwares) files. 
These changes can allow people that record in Nikon microscopes to quickly use Caiman functions without data conversion. Also, it can allow to open big tiff files that come in multiseries, or to save big files iterativelly for later use using the append variable of the tifffile.TiffFile.imwrite() function, which stores bigTiff data in various series.

These changes requires nd2reader as a dependency (accessible through conda-forge)

The changes are:
1 - Creation of a load_multiseries_tiff_file function in caiman.base.movies. This function allows opening any sets of frames based on slice() defined frames. 
2 - Change of load and load_iter functions in caiman.base.movie() to allow opening these two file types 
3 - Change of get _file_size function in caiman.source_extraction.cnmf.utilities to also account for these two new file types.

# Type of change

# Branching
- All PRs should be made against the **dev** branch. The master branch is not often merged back to dev.

# Has your PR been tested?

```python caimanmanager.py test```

and

```python caimanmanager.py demotest```

It passed the demotest and I ran the OnAcid notebook with the original supplied data and with data in these two formats without any problems. However it gave the following error while running the python caimanmanager.py test

error_openmultiseries

### CODE REVIEW

File: caiman/source_extraction/cnmf/utilities.py [hunk 1]
```diff
@@ -33,6 +33,7 @@
 import scipy
 from scipy.sparse import spdiags, issparse, csc_matrix, csr_matrix
 import scipy.ndimage.morphology as morph
+from nd2reader import ND2Reader
```
kushalkolar, line 36: you'd want to make this an optional import, users who don't have nd2 files shouldn't have to install it, something like:

```python

try:
    from nd2reader import ND2Reader
except ImportError:
    HAS_ND2 = False
else:
    HAS_ND2 = True
```

caiman/source_extraction/cnmf/utilities.py
```diff
@@ -992,13 +993,23 @@ def get_file_size(file_name, var_name_hdf5='mov'):
...
-                    T, dims = siz[0], siz[1:]
+                    siz = tffl.series[0].shape
+                    # tiff files written in append mode
+                    if len(siz) < 3:
+                        dims = siz
+                        T = len(tffl.pages)
+                    else:
+                        T, dims = siz[0], siz[1:]
+            elif extension == '.nd2':
+                nd2 = ND2Reader(file_name)
```
kushalkolar, line 1010: from above, something like

```python
if not HAS_ND2:
    raise ImportError("nd2reader must be installed to use nd files")
```

kushalkolar (commented):
I had some comments, but anyways I'm not the maintainer and I'd wait to see what Pat says

caiman/source_extraction/cnmf/utilities.py
pgunn: It's a pure-python dependency available in conda-forge and pypi; having it as a static dependency is fine (preferable actually)

pgunn:
If we're going down the static dependency route (it's a pretty easy dependency I think?) then we don't want any code that checks to see if it's there and things can be simpler.

Instead, please look into modifying the yaml files and the requirements.txt (and I'll modify the conda-forge plumbing to add it as a dependency to the binary package).

Please note though that if this ever in the future does become a troublesome dependency for any reason we'll be pulling it back out of the code.

kushalkolar:
Do you think it's just safer to keep it as an optional dependency?

@RuiOSilva just curious, were your multi series tiff files openable after this PR: https://github.com/flatironinstitute/CaImAn/pull/992

There's a never ending complexity with tiff files :)

pgunn:
> Do you think it's just safer to keep it as an optional dependency?
> 
> @RuiOSilva just curious, were your multi series tiff files openable after this PR: #992
> 
> There's a never ending complexity with tiff files :)

Optional dependencies are something IMO best avoided unless there's a really good reason; they make code analysis hard and make it unsafe to cleanup things. They also confuse static analysis tools.

RuiOSilva:
@kushalkolar yes, these changes allowed me to run multiseries tiff files even with online CNMF, still not with .mmap though.
I will try to do the PR for the motion correction templates tomorrow. I still have some things to change in my codes.

pgunn:
@RuiOSilva Giving you a heads-up that we're going to be resolving the merge conflict and ideally getting this merged into dev in some form over the next few days. Some parts, like the changes to play(), will be too involved to preserve, but the file format work will go in.

pgunn:
Hello,
Apologies for asking you to do this, but we haven't gotten around to looking at this until recently, and in the meantime the load() and load_iter() functions have needed to be reworked pretty substantially. I've been looking through the code trying to find ways to get your load_iter() changes for tiff in there, but it looks like we changed the base code and translating the intent is pretty difficult.

If you wouldn't mind redoing the diff against dev (maybe break it into two diffs, one for nd2reader, one for the tiff changes), ideally not restructuring the code in general too much (add nd2reader to the bottom of the parse list), I'd love to get these changes in there.

Issue #1013, opened 2022-10-13
VolPy Mask_RCNN does not work in CaImAn 1.9.9 or later


KawashimaLab:
- **Tell us a bit about your setup**: 

 1. Linux CentOS7
 
 2. Python 3.9.13, cuda 11, Tesla V100
 
 3. Jupyter Lab (remote)
 
 4. [demo_pipeline_voltage_imaging.py]
 
 5. CaImAn version : 1.9.9 or later
 
 6. mamba and pip

- **Describe the issue that you are experiencing**

We have been using VolPy and it has been working great with Mask_RCNN. We recently tried to upgrade CaImAn to newer versions and met trouble. The installation and most of the functionalities was OK. But MASK_RCNN module does not recognize neurons even in the demo code (0 neurons recognized).

There is no clear error code returned in the script. The code does not stop at MASK_RCNN part. 

We reverted the version to CaImAN to 1.9.8 (Tensorflow 2.7.0) and it works.  CaImAn 1.9.9 and Tensorflow 2.10.0 does not work. We tried to revert Tensorflow to 2.7.0, but couldn't figure out a way in caiman environment. 

Thanks for your help!

Takashi

j-friedrich:
Thanks Takashi for letting us know. I tried, and ran into the same issue using a fresh clone of the current master branch. Changing the environment.yml to force python==3.9 I ended up with tensorflow 2.6 and it works. Oddly, updating tensorflow from 2.6 to 2.7 breaks it for me, whereas you say 2.7 works for you. Don't have a fix yet for this curious issue.

pgunn:
@caichangjia This might be worth digging into

caichangjia:
I will look into that

KawashimaLab:
Thank you Johannes and Changjia! I just double-checked the version. In the current working CaImAn 1.9.8. environment, Python is 3.9.13. Both Tensorflow and Keras are 2.7.0.

caichangjia:
Okay, I finally figure out the problem is because tensorflow changes function to load weights for the neural network in some version. Will ask for pull request soon. @KawashimaLab

KawashimaLab:
Thank you so much!

PR #1067, opened 2023-04-18
Volpy notebook

caichangjia:
# Description
Add ipython notebook for VolPy.


pgunn:
It might be tidier to "Restart and clear output" on the notebook first

EricThomson:
A few minor comments on things probably not needed outside of colab context:

    if __IPYTHON__:
        # this is used for debugging purposes only. allows to reload classes
        # when changed

    
    save the movie in the mp4 format

Note for me it breaks at the `utils.mrcnn_inference` step (it finds no ROIs). I can't see why: at the colab nb it works fine, I seem to have everything installed in my environment. Not sure how much I want to make of this in a PR my guess is if it is an issue people will report it, or I will  once I've had time to properly investigate. That step works fine in colab. I'll have time to look at this more closely next week.

EricThomson:
> if __IPYTHON__:
>     # this is used for debugging purposes only. allows to reload classes
>     # when changed

Oops Nevermind this part is fine. the snippet about saving as mp4 is the only vestigial colab bit I found.

pgunn:
Small nitpick: Does SEGMENTATION need to be in all-caps? (sorry to pester)

caichangjia:
Eric, the VolPy mrcnn inference problem you mentioned is #1013 which should be solved already. I tested on newly downloaded master branch and it worked.

pgunn:
I'm happy to merge; is it ready?

EricThomson:
I think it is ready, this mask-rcnn bugfix was one release ahead of my version of caiman in my env, oops! Sorry @caichangjia thanks for the addition!

PR #1070, opened 2023-04-21
OnACID docs

kushalkolar:
adding some docstrings and comments to OnACID

### DIFF
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -80,6 +80,17 @@ class OnACID(object):
     """
 
     def __init__(self, params=None, estimates=None, path=None, dview=None, Ain=None):
+        """
+
+        Parameters
+        ----------
+        params
+        estimates
+        path
+        dview
+        Ain:
+            binary masked for seeded initialization, used only if "init_method" is set to "seeded"
+        """
         if path is None:
             self.params = CNMFParams() if params is None else params
             self.estimates = Estimates() if estimates is None else estimates
@@ -965,6 +976,8 @@ class OnACID(object):
         self.img_min = img_min
         self.img_norm = img_norm
         if self.params.get('online', 'init_method') == 'bare':
+            # bare is almost like no initialization
+            # cnmf: optional then it runs offline CNMF with offline CNMF init params
             init = self.params.get_group('init').copy()
             is1p = (init['method_init'] == 'corr_pnr' and  init['ring_size_factor'] is not None)
             if is1p:
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -85,7 +85,7 @@ class OnACID(object):
 
         Args:
             params: CNMFParams
-                CNMFParams object with parameters to run online motion correction followed by online CNMF 
+                CNMFParams object with parameters that are used to perform online motion correction, followed by online CNMF 
 
             estimates: Estimates, optional
                 Estimates object to load an existing model
@@ -986,8 +986,8 @@ class OnACID(object):
         self.img_min = img_min
         self.img_norm = img_norm
         if self.params.get('online', 'init_method') == 'bare':
-            # bare is almost like no initialization
-            # cnmf: optional then it runs offline CNMF with offline CNMF init params
+            # bare: no initialization is done
+            # cnmf: runs offline CNMF using offline CNMF init params on a small portion of the movie to obtain spatial footprints of neurons which are then used to seed online CNMF
             init = self.params.get_group('init').copy()
             is1p = (init['method_init'] == 'corr_pnr' and  init['ring_size_factor'] is not None)
             if is1p:
@@ -1080,7 +1080,7 @@ class OnACID(object):
         else:
             raise Exception("Unsupported file extension")
 
-    def mc_next(self, t, frame):
+    def mc_next(self, t: int, frame: np.ndarray):
         """
         Perform online motion correction on the next frame
 
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -91,7 +91,7 @@ class OnACID(object):
                 Estimates object to load an existing model
 
             path: str, optional
-                path to a saved OnACID model on disk
+                path to a saved OnACID model hdf5 file on disk
 
             dview:
                 dview instance
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -85,7 +85,7 @@ class OnACID(object):
 
         Args:
             params: CNMFParams
-                CNMFParams object with parameters for the entire run of initialization and OnACID
+                CNMFParams object with parameters to run online motion correction followed by online CNMF 
 
             estimates: Estimates, optional
                 Estimates object to load an existing model
@@ -94,7 +94,7 @@ class OnACID(object):
                 path to a saved OnACID model hdf5 file on disk
 
             dview:
-                dview instance
+                dview instance, multiprocessing object
 
             Ain: csc_matrix, optional
                 binary masked for seeded initialization as a Compressed Sparse Column matrix.
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -1080,7 +1080,7 @@ class OnACID(object):
         else:
             raise Exception("Unsupported file extension")
 
-    def mc_next(self, t: int, frame: np.ndarray):
+    def mc_next(self, t: int, frame: np.ndarray) -> np.ndarray:
         """
         Perform online motion correction on the next frame
 
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -97,7 +97,8 @@ class OnACID(object):
                 dview instance
 
             Ain: csc_matrix, optional
-                binary masked for seeded initialization, used only if ``"init_method"`` is set to ``"seeded"``
+                binary masked for seeded initialization as a Compressed Sparse Column matrix.
+                To use set ``"init_method"`` to ``"seeded"``
 
         """
         if path is None:
--- a/caiman/source_extraction/cnmf/online_cnmf.py
+++ b/caiman/source_extraction/cnmf/online_cnmf.py
@@ -58,6 +58,7 @@ try:
 except:
     def profile(a): return a
 
+
 class OnACID(object):
     """  Source extraction of streaming data using online matrix factorization.
     The class can be initialized by passing a "params" object for setting up
@@ -82,14 +83,22 @@ class OnACID(object):
     def __init__(self, params=None, estimates=None, path=None, dview=None, Ain=None):
         """
 
-        Parameters
-        ----------
-        params
-        estimates
-        path
-        dview
-        Ain:
-            binary masked for seeded initialization, used only if "init_method" is set to "seeded"
+        Args:
+            params: CNMFParams
+                CNMFParams object with parameters for the entire run of initialization and OnACID
+
+            estimates: Estimates, optional
+                Estimates object to load an existing model
+
+            path: str, optional
+                path to a saved OnACID model on disk
+
+            dview:
+                dview instance
+
+            Ain: csc_matrix, optional
+                binary masked for seeded initialization, used only if ``"init_method"`` is set to ``"seeded"``
+
         """
         if path is None:
             self.params = CNMFParams() if params is None else params
@@ -400,11 +409,11 @@ class OnACID(object):
     def fit_next(self, t, frame_in, num_iters_hals=3):
         """
         This method fits the next frame using the CaImAn online algorithm and
-        updates the object.
+        updates the object. Does NOT perform motion correction, see ``mc_next()``
 
         Args
             t : int
-                time measured in number of frames
+                temporal index of the next frame to fit
 
             frame_in : array
                 flattened array of shape (x * y [ * z],) containing the t-th image.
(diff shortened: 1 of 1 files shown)


caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+            estimates: Estimates, optional
+                Estimates object to load an existing model
+
+            path: str, optional
+                path to a saved OnACID model on disk
+
+            dview:
+                dview instance
+
+            Ain: csc_matrix, optional
```
kushalkolar, line 99: I couldn't find if there's a place where it gets converted to a regular dense array, if a dense array is provided. If you can check and test that it works with dense it could be modified.

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+                binary masked for seeded initialization as a Compressed Sparse Column matrix.
+                To use set ``"init_method"`` to ``"seeded"``
```
kushalkolar, line 101: I'm assuming this is the case

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
     """
 
     def __init__(self, params=None, estimates=None, path=None, dview=None, Ain=None):
+        """
+
+        Args:
+            params: CNMFParams
+                CNMFParams object with parameters for the entire run of initialization and OnACID
+
+            estimates: Estimates, optional
```
kushalkolar, line 90: would probably be useful to state why you might want to construct an `OnACID()` with an existing estimates object? the purpose of this argument isn't clear to me.

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -389,11 +410,11 @@ def get_indices_of_pixels_on_ring(self, pixel):
     def fit_next(self, t, frame_in, num_iters_hals=3):
         """
         This method fits the next frame using the CaImAn online algorithm and
-        updates the object.
+        updates the object. Does NOT perform motion correction, see ``mc_next()``
```
kushalkolar, line 413: I think this is true

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
     """
 
     def __init__(self, params=None, estimates=None, path=None, dview=None, Ain=None):
+        """
+
+        Args:
+            params: CNMFParams
+                CNMFParams object with parameters for the entire run of initialization and OnACID
```
pgunn, line 88: "for the entire run of" is weird phrasing

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+
+            estimates: Estimates, optional
+                Estimates object to load an existing model
+
+            path: str, optional
+                path to a saved OnACID model on disk
```
pgunn, line 94: should we mention what format the model is in? Or point at a doc maybe for how to make on of these?

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+
+            dview:
+                dview instance
```
pgunn, line 97: Needs a bit more explanation

pgunn (commented):
Generally looks good

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: Good point. I'm assuming because the residual buffer, which is used to construct new neural estimates, will be more well behaved if you already have a more robust estimate in hand (A/C/b/f), so this gives you the possibility to pre-set those estimates. E.g., if you recorded for a while and are picking back up after a delay (e.g., you recorded for a while, did some offline analysis or something and now are ready to continue your online analysis). It would be nice to get confirmation from @j-friedrich or @agiovann or @caichangjia on this, as I am speculating.

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: Pat good point., We should probably have an example of this in use in a demo notebook at some poin, given that we want to move toward online as the next gen.

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: I describe it as a "multicore processing object" and think that might be a good generic description?

It would be nice if we had better general documentation on `dview` (in the `setup_cluster()` the docs for it are "dview: ipyparallel dview object, or for multiprocessing: Pool object". But if the users use SLURM I'm not sure what dview will return. I think "multicore processing object" or something similar will work for all use cases?.

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: Regardless, I don't like `csc_matrix`, many people don't know what that means maybe change to `sparse (csc) matrix` so newer people will understand it? I think we should probably do this all over the docs.

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: Right, the logic of `fit_online()` is first `mc_next()` (line 1259), and then `fit_next()` (line 1271).

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: Ths is the type, and in docstrings the type annotation should be presented as-is. There's a description on the next line which explains it more :)

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: "for the full pipeline: initialization -> motion correction -> source extraction"

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: I see one potential use case as serialization, but you would typically use a `@classmethod` for that and not have it in the main constructor which is weird.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: I assuming it's the hdf5 files? Anyways I think the better way to do this (and the estimates above) is to have a `@classmethod`. Such as `OnACID.from_estimates()`, `OnACID.from_hdf()`, etc.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: yea the name dview isn't great and I myself am not sure exactly what this is in different scenarios, feel free to edit/suggest

caiman/source_extraction/cnmf/online_cnmf.py
EricThomson: Yes you are right.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar, line 88:
```suggestion
                CNMFParams object with parameters to run both online motion correct followed by online CNMF 
```

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar, line 94:
```suggestion
                path to a saved OnACID model hdf5 file on disk
```

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+                CNMFParams object with parameters for the entire run of initialization and OnACID
+
+            estimates: Estimates, optional
+                Estimates object to load an existing model
+
+            path: str, optional
+                path to a saved OnACID model hdf5 file on disk
+
+            dview:
+                dview instance
```
kushalkolar, line 97:
```suggestion
                dview instance, multiprocessing object
```

kushalkolar:
let's merge this? OnACID is one of the least documented parts of the codebase and these will help.

pgunn:
Once it's marked no-longer-WIP I'm happy to give it final review

kushalkolar:
Ah right, edited, ready for another look!

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
     """
 
     def __init__(self, params=None, estimates=None, path=None, dview=None, Ain=None):
+        """
+
+        Args:
+            params: CNMFParams
+                CNMFParams object with parameters to run online motion correction followed by online CNMF
```
pgunn, line 88: Feels like a run-on sentence; rephrase this somehow?

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -80,6 +81,26 @@ class OnACID(object):
...
+
+        Args:
+            params: CNMFParams
+                CNMFParams object with parameters to run online motion correction followed by online CNMF 
+
+            estimates: Estimates, optional
+                Estimates object to load an existing model
+
+            path: str, optional
+                path to a saved OnACID model hdf5 file on disk
```
pgunn, line 94: Might be worth linking to a doc (although I don't think we have this yet) on what format this model is in, how to make one, etc etc.. I'd note that as a TODO here

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -965,6 +986,8 @@ def initialize_online(self, model_LN=None, T=None):
         self.img_min = img_min
         self.img_norm = img_norm
         if self.params.get('online', 'init_method') == 'bare':
+            # bare is almost like no initialization
+            # cnmf: optional then it runs offline CNMF with offline CNMF init params
```
pgunn, line 990: I am confused by the online/offline disparity between the self.params.get and the comment here

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -1057,8 +1080,21 @@ def save(self,filename):
         else:
             raise Exception("Unsupported file extension")
 
-
     def mc_next(self, t, frame):
+        """
+        Perform online motion correction on the next frame
+
+        Args:
+            t: int
```
pgunn, line 1088: While you're here, because these types are so simple, mind dropping in the python typing into the prototype for mc_next too? If anything, that's more important than the types being in the comments.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: The docs and nb examples for how to save and load models are very sparse, this is a general thing that needs reworking.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: yea this is confusing, I need to rephrase it. "bare" means no initialization is done at all, "cnmf" means it performs offline CNMF first using the offline CNMF params, and then uses those seeds to perform online CNMF

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -1057,8 +1080,21 @@ def save(self,filename):
         else:
             raise Exception("Unsupported file extension")
 
-
     def mc_next(self, t, frame):
```
kushalkolar, line 1083:
```suggestion
    def mc_next(self, t: int, frame: np.ndarray):
```

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: will do, I find them more readable in the docstring than in the method signature when there are 5+ arguments

caiman/source_extraction/cnmf/online_cnmf.py
pgunn: This is exactly the explanation to put in the comment.

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar, line 88:
```suggestion
                CNMFParams object with parameters that are used to perform online motion correction, followed by online CNMF 
```

caiman/source_extraction/cnmf/online_cnmf.py
pgunn: (it wasn't the "bare" part that I found confusion, it was the online-versus-offline part)

kushalkolar, line 990:
```suggestion
            # bare: no initialization is done
            # cnmf: runs offline CNMF using offline CNMF init params on a small portion of the movie to obtain spatial footprints of neurons which are then used to seed online CNMF
```

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: edited to be clearer

caiman/source_extraction/cnmf/online_cnmf.py
pgunn: Fair, although in such cases it can make sense to split the args out over multiple lines. It's increasingly important to use types in prototypes because IDEs and tools like mypy and pydoc can use types for intelligent things.

pgunn: You could also mark the return type of the function with a type annotation

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: I made #1141 for this

kushalkolar:
edited! :smile:

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -1057,8 +1080,21 @@ def save(self,filename):
         else:
             raise Exception("Unsupported file extension")
 
+    def mc_next(self, t: int, frame: np.ndarray):
```
kushalkolar, line 1083:
```suggestion
    def mc_next(self, t: int, frame: np.ndarray) -> np.ndarray:
```

caiman/source_extraction/cnmf/online_cnmf.py
kushalkolar: Yup I always put them in the signature for that reason too! I meant it's easier to read when the type annotations are in the docstring because then it will render in autodocs, easier to read when it's in the docs because sphinx always squishes together the function signature into one unreadable blob, example: https://caiman.readthedocs.io/en/master/core_functions.html#cnmf

PR #1080, opened 2023-05-11
Embed opencv

j-friedrich:
# Description

add an "embed_opencv" backend to movie.play such that the movie is played within the notebook, not a new window (which leads to a crashing kernel when running e.g. in the cloud)

uses opencv and IPython.display. Surprisingly, (on my MacBook) this is faster than the opencv default of displaying in a separate window

# Type of change

# Has your PR been tested?

Yes
```python caimanmanager.py test```
and
```python caimanmanager.py demotest```

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1187,14 +1187,14 @@ class movie(ts.timeseries):
     def play(self,
              gain: float = 1,
              fr=None,
-             magnification=1,
-             offset=0,
+             magnification: float = 1,
+             offset: float = 0,
              interpolation=cv2.INTER_LINEAR,
              backend: str = 'opencv',
              do_loop: bool = False,
              bord_px=None,
-             q_max=99.75,
-             q_min=1,
+             q_max: float = 99.75,
+             q_min: float = 1,
              plot_text: bool = False,
              save_movie: bool = False,
              opencv_codec: str = 'H264',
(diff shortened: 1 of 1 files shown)


EricThomson:
Tested this today this is a big step forward, you could probably spin this into its own package! One problem is it flickers and the movie sometimes disappears and is replaced with the content of the cells beneath it. Easiest to show what I'm seeing:

https://github.com/flatironinstitute/CaImAn/assets/6643322/fa94ce84-926c-468b-87e0-bf75e4bcbb90

kushalkolar:
I think I know what this is, I've seen it before. It might be trying to clear the output and display in the output area after each frame, and sometimes it's not fast enough so you get that flickering of the output shrinking and resizing.

You shouldn't see it if you use this on the last cell in a nb.

Also, took a quick look into this. It should be possible to add a slider etc. with this type of thing. In fact, it might be possible for me to even wrap fastplotlib `ImageWidget` to target this as a backend. You wouldn't get the other features of fpl of course since this just displays a rasterized jpeg.

Actually, we can just use `jupyter_rfb` since it handles the jpeg encoding etc. and if you're not using pygfx it doesn't need the GPU!

kushalkolar:
What about collab?

kushalkolar:
Getting rid of Qt reliance has other advantages too, such as avoiding conflicts as shown here (I've seen this often on my side too but I don't rely on Qt anymore): https://github.com/flatironinstitute/CaImAn/issues/1064#issuecomment-1538133076

If Qt is no longer necessary the deps can be changed to just require `opencv-python-headless`

PR #1101, opened 2023-06-05
play movies w/o loading them into memory

j-friedrich:
# Description

To play a movie means iterating over its frames and displaying them. If only there was such a thing as a python iterator ;), which there is!

This moves the code of the `movies` `play()` method into a function and adds the option to provide a (list of) filename(s) instead of a `movie` object, which avoids loading the movie in memory by using `load_iter()` instead of `load()`. 

# Type of change

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1240,164 +1240,8 @@ class movie(ts.timeseries):
         Raises:
             Exception 'Unknown backend!'
         """
-        # todo: todocument
-        if backend == 'pylab':
-            logging.warning('*** WARNING *** SPEED MIGHT BE LOW. USE opencv backend if available')
-
-        gain *= 1.     # convert to float in case we were passed an int
-        if q_max < 100:
-            maxmov = np.nanpercentile(self[0:10], q_max)
-        else:
-            maxmov = np.nanmax(self)
-
-        if q_min > 0:
-            minmov = np.nanpercentile(self[0:10], q_min)
-        else:
-            minmov = np.nanmin(self)
-
-        def process_frame(iddxx, frame, bord_px, magnification, interpolation, minmov, maxmov, gain, offset, plot_text):
-            if bord_px is not None and np.sum(bord_px) > 0:
-                frame = frame[bord_px:-bord_px, bord_px:-bord_px]
-            if magnification != 1:
-                frame = cv2.resize(frame, None, fx=magnification, fy=magnification, interpolation=interpolation)
-            frame = (offset + frame - minmov) * gain / (maxmov - minmov)
-
-            if plot_text == True:
-                text_width, text_height = cv2.getTextSize('Frame = ' + str(iddxx),
-                                                            fontFace=5,
-                                                            fontScale=0.8,
-                                                            thickness=1)[0]
-                cv2.putText(frame,
-                            'Frame = ' + str(iddxx),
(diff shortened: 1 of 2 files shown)


caiman/base/movies.py
```diff
@@ -1240,164 +1240,8 @@ def play(self,
...
-                looping = True
-            else:
-                looping = False
-
-        if backend == 'opencv':
-            cv2.waitKey(100)
-            cv2.destroyAllWindows()
-            for i in range(10):
-                cv2.waitKey(100)
+        play(self, gain, fr, magnification, offset, interpolation, backend, do_loop, bord_px,
```
pgunn, line 1243: Confusing to have an object method and a function both called play in the same file; maybe rename the function?

pgunn:
Generally a welcome (and awesome) change!

caiman/base/movies.py
j-friedrich: Should I change it to `play_movie`? 
I noticed `estimates` also has a `play_movie` method, if for now the `embed_opencv` backend is the most convenient to play movies within a notebook, I'll add that backend to `estimates.play_movie` as well.

pgunn: play_movie() would be great.

EricThomson:
This is great I was just discussing this in notebook revision. Testing now will add some comments soon (if I have any), probably tomorrow. 

I'm assuming the changes to `roi.py` are unrelated?

j-friedrich:
> This is great I was just discussing this in notebook revision. Testing now will add some comments soon (if I have any), probably tomorrow.
> 
> I'm assuming the changes to `roi.py` are unrelated?

It just addresses a deprecation warning

EricThomson:
Just tested with standard tif and it works great!  

The only question I have and it is minor: the the downsampling can be helpful for looking at motion artifact. Is there a simple way I'm not seeing to index across the movie here, and get temporal downsampling with this method? E.g., something like the `subindexes` argument to the `load()` functions?

EricThomson:
Very cool! Your points are good: simply downsampling isn't great for SNR as you mentioned, so a `downsample_ratio` would be amazing if not too hard, but even just having indices is very helpful too. I was just trying to think of lower hanging fruit. :smile:

j-friedrich:
Added a `tsub` argument for temporal downscaling. For simplicity of coding, it has to be an integer, which I think suffices. Not sure what the best convention is. `tsub` is already familiar as parameter used to downscale the data (size gets divided by it), whereas `resize` uses a magnification factor of `downscale_factor=1/tsub`.

EricThomson:
> Added a `tsub` argument for temporal downscaling. For simplicity of coding, it has to be an integer, which I think suffices. Not sure what the best convention is. `tsub` is already familiar as parameter used to downscale the data (size gets divided by it), whereas `resize` uses a magnification factor of `downscale_factor=1/tsub`.

Just tested and it works just like the `fz` parameter in `resize()` and looks fantastic.  This will be super useful especially early in the demo notebooks when people just want to quickly inspect their data.

j-friedrich:
Yes, trying to quickly inspect some larger movies prompted me to add this. And in the demos we can get rid of sentences like "Play the movie (optional). This will require loading the movie in memory which in general is not needed by the pipeline."

EricThomson:
Yes, I was drastically revamping that section today and this will help a lot with that (I cut the optional, and was suggesting people use `subindexes` if they run into memory problems, but this is even better).

Issue #1102, opened 2023-06-06
demo_OnACID_mesoscope CLI demo: show_movie is broken


pgunn:
The show_movie flag, when set for the demo_OnACID_mesoscope CLI demo, causes a numpy exception at the end of cnmf/online_cnmf.py:OnACID.initialize_online():

```
        if opts['show_movie']:
            self.bnd_AC = np.percentile(self.estimates.A.dot(self.estimates.C),
                                        (0.001, 100-0.005))
```
The exception is:
```
File "<array_function internals>", line 200, in dot
ValueError: shapes (2,22528000) and (2,1) not aligned: 22528000 (dim 1) != 2 (dim 0)
```

As far as I can tell, this indicates that either the spatial footprints or the temporal traces are not properly initialized.

pgunn:
This happens in the notebook too.

pgunn:
Stepping back through past versions to see if this ever worked.

Caiman 1.9.10 (from about a year) has the problem. (will add further notes as I keep experimenting)

pgunn:
Caiman 1.8.9 works. It must have broken between that version and current.

Installable with:
`mamba create -n caiman-189 -c conda-forge caiman=1.8.9 numpy=1.20.2 scikit-image=0.18`

pgunn:
Caiman 1.9.4 works. Must be between 1.9.4 and 1.9.10.

Installable with:
`mamba create -n caiman-194 -c conda-forge caiman=1.9.4 numpy=1.21.3 scikit-image=0.18.3`

pgunn:
Caiman 1.9.9 works (but seems to display the images a little differently than 1.9.7). Fault probably between 1.9.9 and 1.9.10.

Installable with:
`mamba create -n caiman-199 -c conda-forge caiman=1.9.9 numpy=1.20.3 scikit-image=0.19.2`

pgunn:
Re-verified 1.9.10 doesn't work with a slightly different build instruction:

`mamba create -n caiman-1910v2 -c conda-forge caiman=1.9.10 numpy=1.22.4 scikit-image=0.19.2`

Time to dig into what changed between 1.9.9 and 1.9.10 (both in our code and dependencies)

pgunn:
Verified that it wasn't the move to Python 3.10 in Caiman 1.9.10 (built a 1.9.10 install with python 3.9, which still had the problem)

pgunn:
[...]

PR #1103, opened 2023-06-07
cnmf/online_cnmf.py:OnACID.initialize_online(): Fix call to np.percentile()

pgunn:
numpy 1.22 (over a year ago) completely reworked np.percentile(), causing our old way of calling it to no longer be valid (that way may have never been intended).

Here I flatten the input array down with np.ravel() before passing it in, which seems to give the same result that 1.21 did on the non-flattened array.

I am very unsure about this fix (and we may need to visit other uses of np.percentile() to see if they need adjustment too).


pgunn:
This is meant as a fix for #1102

j-friedrich (approved):
Yes, `np.ravel(...)` is perfect, or `np.array(...)`. The problem is that `A.dot(C)` returns a `matrix`, not an `array`.

EricThomson:
Nice -- very impressive debugging @pgunn !

PR #1109, opened 2023-06-10
Revert "Change doc theme from classic to alabaster"

EricThomson:
Reverts flatironinstitute/CaImAn#1098

There are some problems with API docs so we should figure this out before merging

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -118,7 +118,7 @@ todo_include_todos = False
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'alabaster'
+html_theme = 'classic'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the


kushalkolar:
I think this is because `ipywidgets` is not found, not because of the theme:

https://readthedocs.org/projects/caiman/builds/21013028/

`ipywidgets` was added to `caiman.base.movies` in dev:

https://github.com/flatironinstitute/CaImAn/blob/115fb64915e1e28db6775addf4a6588bb7ec2a01/caiman/base/movies.py#L18

I think if it's added to `docs_require` then docs will build:

https://github.com/flatironinstitute/CaImAn/blob/master/docs/docs_require.txt

BTW, there's an option somewhere to show the docs build as failed if it spits out a warning, like when the API wasn't built. Better to catch it before a user does.

TIL you can lookup the build log of anything on readthedocs!

pgunn:
For a change like this, I'll leave it to you whether to merge it or close it; I have no opinion.

Issue #967, opened 2022-04-10
Component viewers not working as self.R is not None, but b'NoneType'


EricThomson:
1. Operating system (Linux/macOS/Windows): 
 Windows

 2. Python version (3.x): 
 3.9

 3. Working environment (Python IDE/Jupyter Notebook/other):
Jupyter
 
 4. Which of the demo scripts you're using for your analysis (if applicable):
Custom
 
 5. CaImAn version*: 
 1.9.10

 6. CaImAn installation process (`pip install .`/`pip install -e .`/conda): 
pip install -e .

- **Describe the issue that you are experiencing**
Using `nb_view_components()` or `hv_view_components()` I get an error, when I first reconstruct cnmf object after loading from memory using `loadCNMF()` and `estimates.R` is not yet filled (but is not `None` it is `b'NoneType'`).

- **Copy error log below**
```
File d:\caiman\caiman\source_extraction\cnmf\estimates.py:438, in Estimates.hv_view_components(self, Yr, img, idx, denoised_color, cmap)
    436 if self.R is None:
    437     self.R = self.YrA
--> 438 if self.R.shape != [nr, T]:
    439     if self.YrA is None:
    440         self.compute_residuals(Yr)

AttributeError: 'bytes' object has no attribute 'shape'
```

It turns out that `cnm.estimates.R` is not `None` , but `b'NoneType'`, so I fix it by changing line 436 to:

`if self.R is None or self.R == b'NoneType':`

I'm making a PR now, based on this, I will refer to this issue in the PR. I hope that's ok.

PR #1129, opened 2023-07-05
updated vis tools to work with new bokeh

### LINKED ISSUE #1127: Our bokeh calls rely on flags removed in bokeh 3.x
Bokeh 3.x removed plot_width and plot_height:
https://github.com/bokeh/bokeh/wiki/Migration-Guides#changes-310

They suggest using `width` and `height` instead. We should probably review other bokeh changes now that unpinning holoviews pulls several related packages forward into the modern age, graphics stack wise, but we at least need to fix this.

This is a release blocker before the next version.

EricThomson:
# Description

Updating visualization tools to handle updated bokeh as discussed in issue #1127. Basically two problems: the way it handled y origin was changed, and `plot_height` was deprecated. Also, I fixed an unrelated check for `None` that was throwing an error. 

Fixes #1127 

This is not yet finished, as discussed at the issue `nb_view_patches3d()` is still acting weird and I'm not sure how to put in the simple fix. It is probably something really simple I'm missing :smile: 

# Type of change
- [ X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/caiman/source_extraction/cnmf/estimates.py
+++ b/caiman/source_extraction/cnmf/estimates.py
@@ -498,6 +498,8 @@ class Estimates(object):
                 name of colormap (e.g. 'viridis') used to plot image_neurons
 
         """
+        logging.warning("This plotter is likely inaccurate/buggy b/c of a Bokeh update. Fix pending.")
+
         if 'csc_matrix' not in str(type(self.A)):
             self.A = scipy.sparse.csc_matrix(self.A)
         if dims is None:
--- a/caiman/source_extraction/cnmf/estimates.py
+++ b/caiman/source_extraction/cnmf/estimates.py
@@ -282,8 +282,8 @@ class Estimates(object):
                                 self.dims[0], self.dims[1], coordinates=coor_g,
                                 thr_method=thr_method, thr=thr, show=False,
                                 line_color=line_color, cmap=cmap)
-                p1.plot_width = 450
-                p1.plot_height = 450 * self.dims[0] // self.dims[1]
+                p1.width = 450
+                p1.height = 450 * self.dims[0] // self.dims[1]
                 p1.title.text = "Accepted Components"
                 if params is not None:
                     p1.xaxis.axis_label = '''\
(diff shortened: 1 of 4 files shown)


EricThomson:
Currently seems to be working except one relatively minor thing in `nb_view_patches3d()`, when `max_projection` is set to `True`, the column projection appears in the wrong place. I wasn't able to quickly figure out a fix, but I think this is the last obvious problem. 

Here is what it looks like in `max_projection` mode. 

bokeh_3d_view_max

EricThomson:
Oh I guess it was supposed to look like that. Hmm...that's counterintuitive, but ok. I'll resolve conflicts

EricThomson:
Reason for, perhaps counterintuitive, handling of previous:

`if self.R is None or self.R == b'NoneType'`

Is this started to throw an error when `self.R` was an ndarray. The problem is `self.R` is sometimes `b'NoneType'` which is not the same as `None`, so the first condition will not throw a `True`.  The solution I came up with was:

`if self.R is None or not isinstance(self.R, np.ndarray):`

This is related to #967 and #969

j-friedrich:
> Oh I guess it was supposed to look like that. Hmm...that's counterintuitive, but ok. I'll resolve conflicts

Yes, supposed to look like that if you use the current default `axis=0`, I'd call it with `axis=2`, so that the main projection in the top right is along the z-axis not the x-axis. Might want to change the default to 2, assuming z is usually the shortest dimension

EricThomson:
I had tried that, but then panels got ordered correctly, but the image got ordered wrong: 
bokeh_axis2

j-friedrich:
Had the same issue, but fixed it in dev

EricThomson:
Sorry let me look I thought I had the same problem in dev but maybe I used axis=1. let me check.

EricThomson:
Ok right that isn't a problem now. New problem -- you committed directly to dev, while I had already made a PR based on dev, so things are a bit weird. Let me try to make changes you made to dev play nicely with my PR. My fix shifted things back to origin, but yours did too, so now they are shifted up to negative.

EricThomson:
In the `fixing funky visualization` commit, I fixed the problems with the 2d plotters with the most recent changes from the commit clashes, but the 3d has some residual problems. E.g., when plotting layered (no max projection) the image and patches don't seem properly matched : one of them seems flipped on the horizontal axis:

bokeh_layered_view

Then with the max projection view (with `axis=2`) things are very funky, clearly the patches seem shifted, I'm not sure exactly what has gone wrong.

bokeh_max_axis2

When working on this, please work from this branch (even if you have to tear down some of my changes :smile:) as otherwise conflicting commits can get confusing to figure out.

PR #1156, opened 2023-08-19
Adding quilt viewer to visualize how FOV is broken up into patches

### LINKED ISSUE #1116: Add patch visualization for motion correction and CNMF
Motion correction and CNMF/E both operate on patches of images whose width/overlap is determined by user parameters (for motion correction it is set by `overlaps` and `strides`, and for CNMF by `stride` and `rf`).  I have built functionality to visualize these patches superimposed on static image templates:

patch_plotting_example

Especially for motion correction it would be more useful to visualize them superimposed on the actual movies so people can get a better sense of the spatial scales at which piecewise motion correction needs to be performed. 

There are a few ways this could be done (not mutually exclusive):

- I can do this in OpenCV and wrap it into the movie object (we'd feed the stride/overlap params to the play method, maybe when left as `None` patches are not drawn). 
- Fastplotlib would be a natural vehicle for this (@kushalkolar )

I'm currently just tweaking the CNMF demo, and not focusing on motion correction. Hence, I'm going to leave it as a static patch visualization tool for now (probably will do a PR for `utils.visualization.py`).

EricThomson:
# Description

Adds functionality to visualize overlapping patches that split up field of view for motion correction and cnmf for parellelization. Useful when picking different spatial parameters (`rf/stride/gSig/gSiz` for cnmf, and `strides/overlaps` for motion correction).

Thanks to @kushalkolar (mesmerize) for idea of `quilt` name, as `patches` is already used for viewing spatial contours. 

Fixes #1116 

# Type of change

# Has your PR been tested?
I ran the tests there are no new errors introduced in dev branch.  I have tested the plotting functionality fairly extensively, have not yet added new tests to `caiman/tests`

### DIFF
--- a/caiman/utils/visualization.py
+++ b/caiman/utils/visualization.py
@@ -1269,17 +1269,17 @@ def rect_draw(row_minmax: ArrayLike,
               alpha: Optional[float]=0.2, 
               ax: Optional[Any]=None) -> Tuple[Any, Any]:
     """
-    Draw a single transluscent rectangle on given axes object.
+    Draw a single rectangle on given axes object.
     
     Args:
         row_minmax: array-like
             [row_min, row_max] -- upper and lower bounds for rectangle rows (ints)
         col_minmax: array-like 
             [col_min, col_max] -- upper and lower bounds for rectangle columns (ints)
-        color: matplotlib.colors
+        color: matplotlib color
             Any acceptable matplotlib color spec (r,g,b), string, etc., default 'white' 
         alpha : float
-            rectangle alpha (0. to 1., where 1 is opaque), default 0.2
+            opaqueness level (0. to 1., where 1 is opaque), default 0.2
         ax : pyplot.Axes object
             axes object upon which rectangle will be drawn, default None
     
@@ -1322,12 +1322,12 @@ def view_quilt(template_image: np.ndarray,
             row x col summary image upon which to draw patches (e.g., correlation image)
         stride (int) stride between patches in pixels
         overlap (int) overlap between patches in pixels
-        color: matplotlib.colors
+        color: matplotlib color
             Any acceptable matplotlib color (r,g,b), string, etc., default 'white' 
         alpha (float) : patch transparency (0. to 1.: higher is more opaque), default 0.2
         vmin (float) : vmin for plotting underlying template image, default None
         vmax (float) : vmax for plotting underlying template image, default None
-        figsize tuple : fig size in inches (width, height), default (6.,6.)
+        figsize (tuple) : fig size in inches (width, height), default (6.,6.)
     
     Returns:
         ax: pyplot.Axes object
@@ -1341,6 +1341,7 @@ def view_quilt(template_image: np.ndarray,
         
     Note: 
         Currently assumes square patches so takes in a single number for stride/overlap.
+        TODO: implement bokeh version of this function
     """
     im_dims = template_image.shape
     patch_rows, patch_cols = get_rectangle_coords(im_dims, stride, overlap)
@@ -1353,6 +1354,3 @@ def view_quilt(template_image: np.ndarray,
             ax, _ = rect_draw(patch_row, patch_col, color='white', alpha=0.2, ax=ax)
             
     return ax
-
-
-# todo: implement bokeh version of quilt viewer
--- a/caiman/utils/visualization.py
+++ b/caiman/utils/visualization.py
@@ -15,6 +15,7 @@ from IPython.display import HTML
 from math import sqrt, ceil
 import matplotlib as mpl
 import matplotlib.cm as cm
+from matplotlib.patches import Rectangle
 from matplotlib.widgets import Slider
 import numpy as np
 from numpy.typing import ArrayLike
@@ -1228,7 +1229,7 @@ def get_rectangle_coords(im_dims: ArrayLike,
     
     Given dimensions of summary image (rows x colums), stride between patches, and overlap
     between patches, returns row coordinates of the patches in patch_rows, and column 
-    coordinates patches in patch_cols. This is meant to be used by plot_patches().
+    coordinates patches in patch_cols. This is meant to be used by view_quilt().
        
     Args:
         im_dims: array-like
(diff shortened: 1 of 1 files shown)


EricThomson:
Note the natural names for these are patches: in the paper and code we talk about splitting the FOV into patches with stride and overlap. Unfortunately there are already a bunch of patch plotting functions that are for plotting contours of spatial components. Luckily mesmerize had a great name for it: quilt!

I have to admit I just recently turned the corner on type hinting because of @pgunn . A couple of notes: I went ahead and used `Any` for matplotlib graphics objects until support for mypy is a little more mature in matplotlib. E.g., see:

https://github.com/matplotlib/matplotlib/issues/20504
https://github.com/matplotlib/matplotlib/pull/24976

In particular, `rect_draw()` takes in  `plt.Axes`, returns `(plt.Axes, matplotlib.patches.Rectangle)`

For array-like things numpy's `ArrayLike` is great, but currently throws an error when you try to specify types such as `ArrayLike[int]`. There should be support for this eventually (note those issues are worth reading for entertainment value alone):    
https://github.com/beartype/beartype/issues/42#issuecomment-901068836
https://github.com/beartype/beartype/issues/212

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+    """
+    Draw a single transluscent rectangle on given axes object.
+    
+    Args:
+        row_minmax: array-like
+            [row_min, row_max] -- 2-elt int bounds for rect rows
+        col_minmax: array-like 
+            [col_min, col_max] -- int bounds for rect cols
+        color : string
+            rectangle color, default 'yellow'
```
pgunn, line 1279: Can we stick a link in here indicating what colours are valid?

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+        alpha : float
+            rectangle alpha (0. to 1., where 1 is opaque), default 0.3
+        ax : pyplot.Axes object
```
pgunn, line 1282: Eventually we'll want to add proper typing for this, but no rush

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+               overlap: int, 
+               color: Optional[str]='white', 
+               alpha: Optional[float]=0.3, 
+               vmin: Optional[float]=0.0, 
+               vmax: Optional[float]=0.6, 
+               figsize: Optional[Tuple]=(6,6)) -> Any:
+    """
+    Plot patches on template image given stride and overlap parameters. 
+    
+    Given stride and overlap, plot overlapping patch blocks on template image.
```
pgunn, line 1319: Seems to have some redundancy between the 1-sentence and longer description; is this because some tool treats those as separate?

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+    This can be useful for checking motion correction and cnmf spatial parameters. 
+    It ends up looking like a quilt pattern.
+    
+    Args:
+        template_image: ndarray
+            row x col summary image upon which to draw patches (e.g., correlation image)
+        stride (int) stride between patches in pixels
+        overlap (int) overlap between patches in pixels
+        color (str): color of patches, default 'white'
```
pgunn, line 1328: probably no need to mention default when it's visible in params.

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+
+
+def view_quilt(template_image: np.ndarray, 
+               stride: int, 
+               overlap: int, 
+               color: Optional[str]='white', 
+               alpha: Optional[float]=0.3, 
+               vmin: Optional[float]=0.0, 
+               vmax: Optional[float]=0.6, 
+               figsize: Optional[Tuple]=(6,6)) -> Any:
```
pgunn, line 1315: Is this Optional[Tuple[float,float]] or Optional[Tuple[int,int]]?

The default seems to disagree with the comment section

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+            [col_min, col_max] -- int bounds for rect cols
+        color : string
+            rectangle color, default 'yellow'
+        alpha : float
+            rectangle alpha (0. to 1., where 1 is opaque), default 0.3
+        ax : pyplot.Axes object
+            axes object upon which rectangle will be drawn, default None
+    
+    Returns:
+        ax (Axes object)
```
pgunn, line 1286: If we don't want to specify types for these, can we say a little bit more about what we expect them to be? matplotlib.axes?

caiman/utils/visualization.py
```diff
@@ -1217,3 +1218,143 @@ def update(val):
...
+        rect (Rectangle object)
+    """
+    from matplotlib.patches import Rectangle
```
pgunn, line 1289: Imports don't go in methods/functions; please move this to the head of the file

caiman/utils/visualization.py
EricThomson: Good catch. I shouldn't have specified string there. Any Matplotlib color specification, e.g., rgb tuple or list, or string `'red'` is fine (https://matplotlib.org/stable/tutorials/colors/colors.html).  I should probably just put `Any` for this, and be more clear in docs.

caiman/utils/visualization.py
EricThomson: Fixing

caiman/utils/visualization.py
EricThomson: Good catch, fixing (it's float,float)

caiman/utils/visualization.py
EricThomson: Yes this is sort of silly. I can just combine them.

caiman/utils/visualization.py
EricThomson: This is debated all over the place. I've noticed in our code we have it in both places, so I've done this too. But it is a maintenance headache and is asking for trouble/inconsistencyes. Happy to stop doing it moving forward.

pgunn: It's not really a big deal either way, it's just another thing to maintain if it changes in one place.

Issue #1155, opened 2023-08-18
Proposal: Implement lazy computing arrays for CNMF outputs


kushalkolar:
Visualization of the reconstructed movie `A * C`, reconstruted background `b * f`, and residuals `Y - (A * C) - (b * f)` are extremely useful for various things like compoment evaluation, quality evaluation, parameter evaluation, etc. These dense arrays are often too large to fit into RAM and we receive questions from users both here and on gitter about this quite often. I propose implementing these using lazy arrays that perform computation on the fly. This is already implemented in [`mesmerize-core`](https://github.com/nel-lab/mesmerize-core/blob/master/mesmerize_core/arrays/_cnmf.py), so I propose just moving that implementation here. The only dependencies are `numpy` and `scipy` which `caiman` already has. We could alternatively implement it using `dask`, but that would add a large dependency.

pgunn:
How does this approach compare to fully computing the arrays and having them reside on disk, instead lazy-loading them? Why is computing them lazily preferable?

kushalkolar:
The full dense array is often hundreds of GB. And it's just not necessary to have the full array, even for things like max projections which can be calculated without the full array because its an outer product.

EricThomson:
This is typically be used for plotting functions in mesmerize, right? Is that the main use case? 

E.g., when plotting a derived quantity (e.g., A*C) you don't want to (and sometimes can't) load into memory, it would be a waste to save those matrix products as memmapped files, but you _can_ calculate the bits you currently are trying to view "on the fly" easily using indexing (even if the individual matrices are memmapped) using lazy computing.  Is the main payoff with dask that it gives you blocks?

pgunn:
How trivial are the computations? I get the storage benefits (and they could be pretty substantial)

With lazy-loading, the data is essentially read-only when done, which is good for data provenance and presumably means all the compute costs are already paid for any subsequent analysis, and it's going to be potentially compatible with other software that might want to load the format.

Maybe none of this is a concern if the lazy computation really is that deterministic and it doesn't get into the main caiman algorithms -particularly anything like setup_cluster()

pgunn:
I think I misunderstood the proposal on first read; assuming what you meant was exactly the computations above, this seems reasonable, although is there a reason to have the arrays themselves do it rather than have explicit code do the needed computations piecewise?

[...]

PR #1167, opened 2023-09-06
Lazy arrays for CNMF

kushalkolar:
Implements #1155 

Adds:

1.`LazyArray` base class which has relevent parts of the numpy ndarray structure to behave like a numpy array. 
2. `LazyArrayRCM`, for the reconstructed movie ``A * C``. 
3. `LazyArrayRCB`, for the reconstructed background ``b * f``. This is actually just a sublcass of `LazyArrayRCM` to separate them. I am not sure if it's really necessary, we could have one class `LazyDenseMovie` or something which works for both.
4. `LazyArrayResiduals`

These are very useful because they also compute projection images without computing the full dense array. 

These should work for visuaslization in most tools that take numpy-like array. It should work in Johannes' ipycanvas based player, and it's already known to work in `fastplotlib` with random access slicing. 

Thoughts:
1. Do we want to support saving the dense array to disk, such as in hdf5 which can be lazily accessed? I'm not sure how useful this could be because the arrays can be huge in the hundreds of gigabytes range

2. I have never worked with 3D data, but this shouldn't be difficult to expand to 3D. Johannes might be the right person to ask? 

main todo:
- [x] min max and projection images for residuals 
- [ ] tests, they already exist in mesmerize-core
- [ ] add to demo nbs

Thoughts and a big-picture review would be useful before a more detailed review.

### DIFF
--- a/caiman/source_extraction/cnmf/lazy_arrays.py
+++ b/caiman/source_extraction/cnmf/lazy_arrays.py
@@ -1,14 +1,16 @@
-from warnings import warn
+#!/usr/bin/env
+
 from abc import ABC, abstractmethod
-from pathlib import Path
-from typing import *
 from itertools import product as iter_product
+from pathlib import Path
 from time import time
+from typing import *
+from warnings import warn
 
 
+import h5py
 import numpy as np
 from scipy.sparse import csc_matrix
-import h5py
 
 
 # TODO: Is there a better way to do this?
@@ -84,7 +86,7 @@ class LazyArray(ABC):
     @abstractmethod
     def _compute_at_indices(self, indices: Union[int, slice]) -> np.ndarray:
         """
-        Lazy computation logic goes here. Computes the array at the desired indices.
+        Must be implemented in sublcass. Lazy computation logic goes here. Computes the array at the desired indices.
 
         Args:
             indices (Union[int, slice]): the user's desired slice, i.e. slice object or int passed from `__getitem__()`
@@ -161,7 +163,7 @@ class LazyArray(ABC):
 
         else:
             raise IndexError(
-                f"You can index LazyArrays only using slice, int, or tuple of slice and int, "
+                f"You can index LazyArrays only using slice, range, int, or tuple of slice and int, "
                 f"you have passed a: <{type(item)}>"
             )
 
@@ -273,10 +275,10 @@ class LazyArrayRCM(LazyArray):
         temporal_std = np.nanstd(self.temporal, axis=1)
 
         # compute mean, max, min, and std projection images
-        self._mean_image = self.spatial.dot(temporal_mean).reshape(frame_dims, order="F")
-        self._max_image = self.spatial.dot(temporal_max).reshape(frame_dims, order="F")
-        self._min_image = self.spatial.dot(temporal_min).reshape(frame_dims, order="F")
-        self._std_image = self.spatial.dot(temporal_std).reshape(frame_dims, order="F")
+        self._mean_image = (self.spatial @ temporal_mean).reshape(frame_dims, order="F")
+        self._max_image = (self.spatial @ temporal_max).reshape(frame_dims, order="F")
+        self._min_image = (self.spatial @ temporal_min).reshape(frame_dims, order="F")
+        self._std_image = (self.spatial @ temporal_std).reshape(frame_dims, order="F")
 
     @property
     def spatial(self) -> np.ndarray:
@@ -327,9 +329,7 @@ class LazyArrayRCM(LazyArray):
         return self._std_image
 
     def _compute_at_indices(self, indices: Union[int, Tuple[int, int]]) -> np.ndarray:
-        rcm = self.spatial.dot(
-            self.temporal[:, indices]
-        ).reshape(
+        rcm = (self.spatial @ self.temporal[:, indices]).reshape(
             self.shape[1:] + (-1,), order="F"
         ).transpose([2, 0, 1])
 
@@ -344,8 +344,8 @@ class LazyArrayRCM(LazyArray):
                f"n_components: {self.n_components}"
 
     def __eq__(self, other):
-        if not isinstance(other, LazyArrayRCM):
-            raise TypeError(f"cannot compute equality for against types that are not {self.__class__.__name__}")
+        if not isinstance(other, self.__class__):
+            raise TypeError(f"cannot compute equality against types that are not {self.__class__.__name__}")
 
         if (self.spatial == other.spatial) and (self.temporal == other.temporal):
             return True
@@ -368,6 +368,7 @@ class LazyArrayResiduals(LazyArray):
             raw_movie: np.memmap,
             rcm: LazyArrayRCM,
             rcb: LazyArrayRCB,
+            timeout: float = 10.0,
     ):
         """
         Construct a LazyArray of the residuals, ``Y - (A ⊗ C) - (b ⊗ f)``
@@ -379,6 +380,8 @@ class LazyArrayResiduals(LazyArray):
 
             rcb (LazyArrayRCB): ``b ⊗ f``, reconstructed background lazy array
 
+            timeout (float): number of seconds allowed for min max calculation of raw video
+
         """
         self._raw_movie = raw_movie
         self._rcm = rcm
@@ -388,29 +391,31 @@ class LazyArrayResiduals(LazyArray):
         self._shape = self._raw_movie.shape
 
         # TODO: implement min max for residuals
-        # min_max_raw = self._quick_min_max(raw_movie, timeout)
-        # if min_max_raw is None:
-        #     self._min = None
-        #     self._max = None
+        min_max_raw = self._quick_min_max(raw_movie, timeout)
+        if min_max_raw is None:
+            self._min = None
+            self._max = None
 
-        # else:
-        #     _min, _max = min_max_raw
-        #
-        #     _min = _min - self._rcm.max - self._rcb.max
-        #     _max = _max -
+        else:
+            _min, _max = min_max_raw
+
+            _min = _min - self._rcm.max - self._rcb.max
+            _max = _max + self._rcm.max + self._rcb.max
+
+
 
     def _quick_min_max(self, data, timeout):
         # adapted from pyqtgraph.ImageView
         # Estimate the min/max values of *data* by subsampling.
         # Returns [(min, max), ...] with one item per channel
 
-        t = time()
+        t0 = time()
         while data.size > 1e6:
             ax = np.argmax(data.shape)
             sl = [slice(None)] * data.ndim
             sl[ax] = slice(None, None, 2)
             data = data[tuple(sl)]
-            if (time() - t) > timeout:
+            if (time() - t0) > timeout:
                 return None
 
         return float(np.nanmin(data)), float(np.nanmax(data))
--- a/caiman/source_extraction/cnmf/lazy_arrays.py
+++ b/caiman/source_extraction/cnmf/lazy_arrays.py
@@ -368,10 +368,14 @@ class LazyArrayResiduals(LazyArray):
             raw_movie: np.memmap,
             rcm: LazyArrayRCM,
             rcb: LazyArrayRCB,
-            timeout: float = 10.0,
+            raw_movie_mean_projection: Optional[np.ndarray] = None,
+            raw_movie_max_projection: Optional[np.ndarray] = None,
+            raw_movie_min_projection: Optional[np.ndarray] = None,
     ):
         """
-        Construct a LazyArray of the residuals, ``Y - (A ⊗ C) - (b ⊗ f)``
+        Construct a LazyArray of the residuals, ``Y - (A ⊗ C) - (b ⊗ f)``.
+
+        If projections of the raw movie are provided, then projections of the residuals will also be computed.
 
         Args:
             raw_movie (np.memmap): ``Y``, numpy memmap of the raw movie
@@ -382,6 +386,12 @@ class LazyArrayResiduals(LazyArray):
 
             timeout (float): number of seconds allowed for min max calculation of raw video
 
+            raw_movie_mean_projection (np.ndarray): optional, mean projection of the raw movie
+
+            raw_movie_max_projection (np.ndarray): optional, max projection of the raw movie
+
+            raw_movie_min_projection (np.ndarray): optional, min projection of the raw movie
+
         """
         self._raw_movie = raw_movie
         self._rcm = rcm
@@ -390,35 +400,16 @@ class LazyArrayResiduals(LazyArray):
         # shape of the residuals will be the same as the shape of the raw movie
         self._shape = self._raw_movie.shape
 
-        # TODO: implement min max for residuals
-        min_max_raw = self._quick_min_max(raw_movie, timeout)
-        if min_max_raw is None:
-            self._min = None
-            self._max = None
+        if raw_movie_mean_projection is not None:
+            self._mean_image = raw_movie_mean_projection - self._rcm.mean_image - self._rcb.mean_image
 
-        else:
-            _min, _max = min_max_raw
+        if raw_movie_max_projection is not None:
+            self._max_image = raw_movie_max_projection - self._rcm.max_image - self._rcb.max_image
+            self._max = raw_movie_max_projection.max() - self._rcm.max - self._rcb.max
 
-            _min = _min - self._rcm.max - self._rcb.max
-            _max = _max + self._rcm.max + self._rcb.max
-
-
-
-    def _quick_min_max(self, data, timeout):
-        # adapted from pyqtgraph.ImageView
-        # Estimate the min/max values of *data* by subsampling.
-        # Returns [(min, max), ...] with one item per channel
-
-        t0 = time()
-        while data.size > 1e6:
-            ax = np.argmax(data.shape)
-            sl = [slice(None)] * data.ndim
-            sl[ax] = slice(None, None, 2)
-            data = data[tuple(sl)]
-            if (time() - t0) > timeout:
-                return None
-
-        return float(np.nanmin(data)), float(np.nanmax(data))
+        if raw_movie_min_projection is not None:
+            self._min_image = raw_movie_max_projection - self._rcm.min_image - self._rcb.min_image
+            self._min = raw_movie_min_projection.min() - self._rcm.min - self._rcb.min
 
     @property
     def shape(self) -> Tuple[int, int, int]:
@@ -428,16 +419,25 @@ class LazyArrayResiduals(LazyArray):
     def n_frames(self) -> int:
         return self._shape[0]
 
-    # TODO: implement min max for residuals
     @property
     def min(self) -> float:
-        raise NotImplementedError
+        return self._min
 
     @property
     def max(self) -> float:
-        raise NotImplementedError
+        return self._max
 
-    # TODO: implement projection images for residuals
+    @property
+    def mean_image(self) -> np.ndarray:
+        return self._mean_image
+
+    @property
+    def max_image(self) -> np.ndarray:
+        return self._max_image
+
+    @property
+    def min_image(self) -> np.ndarray:
+        return self._min_image
 
     def _compute_at_indices(self, indices: Union[int, slice]) -> np.ndarray:
         residuals = self._raw_movie[indices] - self._rcm[indices] - self._rcb[indices]
--- a/caiman/source_extraction/cnmf/lazy_arrays.py
+++ b/caiman/source_extraction/cnmf/lazy_arrays.py
@@ -1,0 +1,439 @@
+from warnings import warn
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import *
+from itertools import product as iter_product
+from time import time
+
+
+import numpy as np
+from scipy.sparse import csc_matrix
+import h5py
+
+
+# TODO: Is there a better way to do this?
+slice_or_int_or_range = Union[int, slice, range]
+
+
+class LazyArray(ABC):
+    """
+    Base class for arrays that exhibit lazy computation upon indexing
+    """
+    @property
+    def dtype(self) -> np.dtype:
+        """
+        np.dtype: numpy dtype
+        """
+        # compute the slice at the first index and return the dtype of that
+        # TODO: there might be a nicer way to do this
+        return self[0].dtype
+
+    @property
+    @abstractmethod
+    def shape(self) -> Tuple[int, int, int]:
+        """
+        Tuple[int]: (n_frames, dims_x, dims_y)
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def n_frames(self) -> int:
+        """
+        int: number of frames
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def min(self) -> float:
+        """
+        float: min value of the array if it were fully computed
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def max(self) -> float:
+        """
+        float: max value of the array if it were fully computed
+        """
+        pass
+
+    @property
+    def ndim(self) -> int:
+        """
+        int: Number of dimensions
+        """
+        return len(self.shape)
+
+    @property
+    def nbytes(self) -> int:
+        """
+        int: number of bytes required for the array if it were fully computed
+        """
+        return np.prod(self.shape + (np.dtype(self.dtype).itemsize,))
+
+    @property
+    def nbytes_gb(self) -> float:
+        """
+        float: number of gigabytes required for the array if it were fully computed
+        """
+        return self.nbytes / 1e9
+
+    @abstractmethod
+    def _compute_at_indices(self, indices: Union[int, slice]) -> np.ndarray:
+        """
+        Lazy computation logic goes here. Computes the array at the desired indices.
+
+        Args:
+            indices (Union[int, slice]): the user's desired slice, i.e. slice object or int passed from `__getitem__()`
+
+        Returns:
+            np.ndarray: slice of the array at the desired indices
+
+        """
+        pass
+
+    def as_numpy(self):
+        """
+        Converts to a standard numpy array in RAM.
+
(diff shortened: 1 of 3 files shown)


caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import *
+from itertools import product as iter_product
+from time import time
+
+
+import numpy as np
+from scipy.sparse import csc_matrix
+import h5py
```
pgunn, line 11: If we can alphabetise the imports and add a #!/usr/bin/env python at the top to hint to editors that it's python

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+# TODO: Is there a better way to do this?
+slice_or_int_or_range = Union[int, slice, range]
```
pgunn, line 15: Wondering what the convention is; do people normally use titlecase for types like this?

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+    def nbytes_gb(self) -> float:
+        """
+        float: number of gigabytes required for the array if it were fully computed
+        """
+        return self.nbytes / 1e9
+
+    @abstractmethod
+    def _compute_at_indices(self, indices: Union[int, slice]) -> np.ndarray:
+        """
+        Lazy computation logic goes here. Computes the array at the desired indices.
```
pgunn, line 87: Is this leaving room for future work?

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+        size of the full dense array**
+
+        Args:
+            filename:
+            dataset_name:
+
+        Returns:
+
+        """
+        pass
```
pgunn, line 135: presumably this is more WIP

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+                    f"Cannot index more dimensions than exist in the array. "
+                    f"You have tried to index with <{len(item)}> dimensions, "
+                    f"only <{len(self.shape)}> dimensions exist in the array"
+                )
+
+            indexer = item[0]
+
+        else:
+            raise IndexError(
+                f"You can index LazyArrays only using slice, int, or tuple of slice and int, "
```
pgunn, line 164: or range

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+            return self._compute_at_indices(indexer)
+
+    def __repr__(self):
+        return f"{self.__class__.__name__} @{hex(id(self))}\n" \
+               f"{self.__class__.__doc__}\n" \
+               f"Frames are computed only upon indexing\n" \
+               f"shape [frames, x, y]: {self.shape}\n"
+
+
+class LazyArrayRCM(LazyArray):
```
pgunn, line 214: I wonder if spelling it out is preferable. It's already kinda long though; LazyArrayReconstructedMovie might be a bit ridiculous.

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+            return rcm
+
+    def __repr__(self):
+        r = super().__repr__()
+        return f"{r}" \
+               f"n_components: {self.n_components}"
+
+    def __eq__(self, other):
+        if not isinstance(other, LazyArrayRCM):
+            raise TypeError(f"cannot compute equality for against types that are not {self.__class__.__name__}")
```
pgunn, line 348: "for against types"?

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,439 @@
...
+        if (self.spatial == other.spatial) and (self.temporal == other.temporal):
+            return True
+        else:
+            return False
+
+
+# implementation for reconstructed background is identical
+# this is just a subclass to separate them
+# TODO: Is this really necessary? Maybe we just make one class called "Reconstructed Array"
+#  and they keep track of spatial and temporal components that the construct them with?
```
pgunn, line 359: If it can be made generic without loss of functionality or clarity that'd be good, but if not, prefer clarity.

"that they construct them with"

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: this is just an abstract method indicating that it must be implemented in the subclass to handle the lazy compute logic for the given slice indices.

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: yes I am not sure how useful this really would be since it can be hundreds of gigabytes, thoughts @EricThomson  ?

pgunn: We can probably skip as_numpy() but save_hdf5 might be useful (provided we make sure it's also not a ram gobbler), particularly for smaller datasets.

pgunn: Actually, maybe zarr or n5 would be a better choice

kushalkolar: Yes! `zarr` would be the natural choice, though we would need to add that as a dependency?

pgunn: I ordinarily hate to have dynamic imports; z5py is the sole example of this in our codebase (to my knowledge) because it's not available in pip (for the few cases we care about pip). I don't mind our code generally depending on it and saying "you're missing a few features" to the people with pip-based installs (they're already off the beaten path).

kushalkolar: `zarr` is on both conda-forge and pypi:
https://anaconda.org/conda-forge/zarr
https://pypi.org/project/zarr/

I wonder if it's cross-compatible with `z5py`

EricThomson:
It will probably take me a while to really think about this properly in the weeds, but big picture thoughts for now (also forgive if this is somewhat terse I'm jet lagged and have to rush to something soon :smile: ):
1) Does this belong in cnmf or utils or visualization? Is this mainly an aid for visualizating? If so, should it go with utils, vis? Speaking generally, isn't this more generally applicable than cnmf, which is a folder fairly explicitly reserved for the cnmf algorithm machinery, while other stuff goes elsewhere. So (again, thinking big picture modular decomposition here), where should this really go? I'm thinking not the cnmf dir as this can also be really useful ultimately for motion correction visualizations and things like that (and future decompositions if we provide them). 
2)  I don't love the naming: RCM /RCB are names people won't understand by looking at them. We should think of more transparent names. I know 'reconstructed movie' and reconstructed background are probably too long to write out fully, but rcb and rcm are abbreviations, and I prefer to not have full blown abbreviations that people will have to learn.
3) Background will ultimately be general not just CNMF, right is still the bg model (i.e., need b0 and W for CNMFE).  This is probably important to get updated on first pass.
4)  I would prefer to use more numpythonic `@` for matrix multiplication. Last time I checked there was no difference, so I think `@` is better/more transparent than `dot` which will tend to confuse people who will think "Why are they doing a dot product rather than matrix multiplication?" As time goes on this will get more confusing to modern readers. I know caiman has `dot` all over the place but that's because it is basically legacy linear algebra code from before numpy changed the notation, and we should just update it everywhere at some point. [Hackathon! :smile: ]

These are just my big picture thoughts for now I will have time to go into the weeds next week.

kushalkolar:
> It will probably take me a while to really think about this properly in the weeds, but big picture thoughts for now (also forgive if this is somewhat terse I'm jet lagged and have to rush to something soon 😄 ):
> 
> 1. Does this belong in cnmf or utils or visualization? Is this mainly an aid for visualizating? If so, should it go with utils, vis? Speaking generally, isn't this more generally applicable than cnmf, which is a folder fairly explicitly reserved for the cnmf algorithm machinery, while other stuff goes elsewhere. So (again, thinking big picture modular decomposition here), where should this really go? I'm thinking not the cnmf dir as this can also be really useful ultimately for motion correction visualizations and things like that (and future decompositions if we provide them).

This is not a visualization aide, it's a back-end data array structure that makes larger-than-RAM array visualization possible. You can use it for calculations beyond just visualization. This is very CNMF specific and not useful and comletely unrelated to motion correction visualization. 

> 2. I don't love the naming: RCM /RCB are names people won't understand by looking at them. We should think of more transparent names. I know 'reconstructed movie' and reconstructed background are probably too long to write out fully, but rcb and rcm are abbreviations, and I prefer to not have full blown abbreviations that people will have to learn.

Any ideas? 

> 3. Background will ultimately be general not just CNMF, right ? Currently bg is just CNMF but we'd need CMFE added (i.e., need b0 and W for CNMFE), right?  This is probably important to get updated on first pass.

Right, I will figure this out. 

> 4. I would prefer to use more numpythonic `@` for matrix multiplication. Last time I checked there was no difference in terms of profiling, but semantically I think `@` is better/more transparent than `dot` which will tend to confuse people who will think "Why are they doing a dot product rather than matrix multiplication?" As time goes on this will get more confusing to modern readers. I know caiman has `dot` all over the place but that's because it is basically legacy linear algebra code from before numpy changed the notation, and we should just update it everywhere at some point. [Hackathon! 😄 ]

I tested this and yes it's the same between `dot` and `@`. BUT `np.outer()` behaves weird for some reason. 

image

kushalkolar:
So for `LazyArrayResiduals` all image projections except std are implemented. Computing `std(Y - (A* C) - (b * f))` is non-trivial, this is probably best handled with dask so we can hopefully have this in the future!
- Basically: `std(Y) - std(A * C) - std(b * f)` != `std(Y - (A* C) - (b * f))` because `sqrt(x - y) != sqrt(x) - sqrt(y)` 

@EricThomson we will need to chat sometime about `b0` and `W`, I'm not familiar with it :D

kushalkolar:
This shows how useful the residuals projection can be, left is K = 1, right is K = 4, you can see neurons left over in the residuals projection. 

image

kushalkolar:
I should make the `@min`, `@max` as methods and it actually should be possible to get min, max across an axis. Implement nanmin(), nanmax() as just aliases to min(), max().

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: I like all these options. :guitar:

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: On the naming options: as I mentioned in my general comment, I am not a fan of the abbreviations RCM etc. I think we should probably cut the mention of array in the names, and that they are lazy. E.g., we could just do `ReconstructedMovie`, `ReconstructedBackground`, `ReconstructedResiduals`. Refer to the objects by what they are, not their data type or back end details -- we don't name zarr arrays in terms of their ability to be randomly accessed.  Yes, these details are obviously really important for our goals, but once these things have inherited from `LazyArray()`, do we want our naming conventions to be captive to this?

EricThomson:
> So for `LazyArrayResiduals` all image projections except std are implemented. Computing `std(Y - (A* C) - (b * f))` is non-trivial, this is probably best handled with dask so we can hopefully have this in the future!
> 
>     * Basically standard deviation is not distributive: `std(Y) - std(A * C) - std(b * f)` != `std(Y - (A* C) - (b * f))` because `sqrt(x - y) != sqrt(x) - sqrt(y)`
> 
>     * The rest of the operations are distributive which is why we can exploit these tricks :D
> 
> 
> @EricThomson we will need to chat sometime about `b0` and `W`, I'm not familiar with it :D

I just found two methods in `cnm.estimates` I hadn't seen before: `compute_background()` and `compute_residuals()` and both handle the ring model and all its funky tributaries nicely.

caiman/source_extraction/cnmf/lazy_arrays.py
pgunn: Let's try to find a name that's clear but not so long. RCM may be too short, but ReconstructedMovie feels long enough to be Java-ish.

pgunn:
> 4. I would prefer to use more numpythonic `@` for matrix multiplication. Last time I checked there was no difference in terms of profiling, but semantically I think `@` is better/more transparent than `dot` which will tend to confuse people who will think "Why are they doing a dot product rather than matrix multiplication?" As time goes on this will get more confusing to modern readers. I know caiman has `dot` all over the place but that's because it is basically legacy linear algebra code from before numpy changed the notation, and we should just update it everywhere at some point. [Hackathon! 😄 ]

If we think it should be done, let's just do it. Converting all instances of .dot() to @ is drudgework; not really suitable for a hackathon.

kushalkolar:
> > 4. I would prefer to use more numpythonic `@` for matrix multiplication. Last time I checked there was no difference in terms of profiling, but semantically I think `@` is better/more transparent than `dot` which will tend to confuse people who will think "Why are they doing a dot product rather than matrix multiplication?" As time goes on this will get more confusing to modern readers. I know caiman has `dot` all over the place but that's because it is basically legacy linear algebra code from before numpy changed the notation, and we should just update it everywhere at some point. [Hackathon! 😄 ]
> 
> If we think it should be done, let's just do it. Converting all instances of .dot() to @ is drudgework; not really suitable for a hackathon. I can probably do it all with regexes without taking much time. Maybe even yapf can do it for us, or some other tool.

Not _every_ place where `.dot()` is used is meant to be a matrix multiplication so it would be dangerous to do it like that. And without tests for all the linalg in the codebase I wouldn't do it. There is nothing that we gain from changing every `.dot()` to `@` where it's possible. I think most people familiar with `numpy` will know that `.dot()` and `@` behave the same when it comes to matrix multiplication.

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: These are well within the bounds of what I'd be comfortable with lengthwise, but if anyone can think of apt shorter names I'd be fine if they don't have hard to understand abbreviations. Also, classes are typically not instantiated that much, and people can name the instances shorter little nubs as they like.

pgunn: I'm not happy with those lengths. If you ever look at code in Java it has those kinds of long names and there is a lot of typing.

Maybe ReconstMov, ReconstBkg, ReconstResids?

Maybe even RcMov, RcBkg, RcResids?

kushalkolar:
So I guess we've arrived at `ModelMovie`, `MovelBackground`, `ModelResiduals` ?

EricThomson:
> So I guess we've arrived at `ModelMovie`, `MovelBackground`, `ModelResiduals` ?
> 
> @EricThomson

I'm ok with that or `Reconstructed` or `Estimated`. In my opinion -- you pick, ultimately any of them work :smile:

pgunn:
> > So I guess we've arrived at `ModelMovie`, `MovelBackground`, `ModelResiduals` ?
> > @EricThomson
> 
> I'm ok with that or `Reconstructed` or `Estimated`. In my opinion -- you pick, ultimately any of them work 😄

Model is shorter; that's the right choice unless we can find something similarly short that we like better.

EricThomson:
> So I guess we've arrived at `ModelMovie`, `MovelBackground`, `ModelResiduals` ?
> 
> @EricThomson

Thinking about this more: one option of the three parts is `NeuralModel` (this is `AC` right?), `BackgroundModel` (this is going to be `bf` or the ring model of the background), and `Residual` (this is basically `Y-NeuralModel-BackgroundModel`, so is basically the error term, all the stuff the model doesn't capture).

pgunn:
> > So I guess we've arrived at `ModelMovie`, `MovelBackground`, `ModelResiduals` ?
> > @EricThomson
> 
> Thinking about this more: one option of the three parts is `NeuralModel` (this is `AC` right?), `BackgroundModel` (this is going to be `bf` or the ring model of the background), and `Residual` (this is basically `Y-NeuralModel-BackgroundModel`, so is basically the error term, all the stuff the model doesn't capture, so isn't really technically part of the model at all).
> 
> Or even "movie" . So something like `NeuralMovie`, `BackgroundMovie`, and `ResidualMovie` (maybe bad, just spitballing again).

I'm not a huge fan of Movie, but Caiman already uses the term so maybe we're committed to it. I wonder if an interesting first term might be Focus, e.g. MovieFocus, MovieBackg, MovieResids

pgunn:
> Actually good point that these could get confused with caiman movie object, and one reason I didn't like it is that it is slipping back to naming metadata (what it is) rather than what type of information it contains. That is, the info the three terms contain is the model's predictions of neural activity, the model's predictions of background activity, and the residual (i.e., the stuff the model leaves out, the error term).

On that narrow point, I don't think it's bad (at all) to name classes partly by implementation details (provided the name doesn't get too long). It's super common in programming to do so ; the distinction between an int and a long is entirely implementation details even though both are meant to hold numbers. There are commonly a few ways to represent something anyhow, and they often should have distinct names based on the how.

EricThomson:
> > Actually good point that these could get confused with caiman movie object, and one reason I didn't like it is that it is slipping back to naming metadata (what it is) rather than what type of information it contains. That is, the info the three terms contain is the model's predictions of neural activity, the model's predictions of background activity, and the residual (i.e., the stuff the model leaves out, the error term).
> 
> On that narrow point, I don't think it's bad (at all) to name classes partly by implementation details (provided the name doesn't get too long). It's super common in programming to do so ; the distinction between an int and a long is entirely implementation details even though both are meant to hold numbers. There are commonly a few ways to represent something anyhow, and they often should have distinct names based on the how.

Yes, but here we are arguing about space so want to eliminate words, and in Python it is considered a code smell to name variables with type flags e.g., `data_float`. But foolish consistency and all that -- it can sometimes be useful.

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def __repr__(self):
+        return f"{self.__class__.__name__} @{hex(id(self))}\n" \
+               f"{self.__class__.__doc__}\n" \
+               f"Frames are computed only upon indexing\n" \
+               f"shape [frames, x, y]: {self.shape}\n"
+
+
+class LazyArrayRCM(LazyArray):
+    """LazyArray for reconstructed movie, i.e. A ⊗ C"""
```
EricThomson, line 217: Maybe just AC  (that ⊗ symbol is usually for outer product)

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,444 @@
...
+        else:
+            return False
+
+
+# implementation for reconstructed background is identical
+# this is just a subclass to separate them
+# TODO: Is this really necessary? Maybe we just make one class called "Reconstructed Array"
+#  and they keep track of spatial and temporal components that the construct them with?
+class LazyArrayRCB(LazyArrayRCM):
+    """Lazy array for reconstructed background, i.e. b ⊗ f"""
```
EricThomson, line 361: Similar minor thing with ⊗ here. It's lazily extracting segments of bf  (basically everywhere insert this :smile:)

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,444 @@
...
+        if not isinstance(other, self.__class__):
+            raise TypeError(f"cannot compute equality against types that are not {self.__class__.__name__}")
+
+        if (self.spatial == other.spatial) and (self.temporal == other.temporal):
+            return True
+        else:
+            return False
+
+
+# implementation for reconstructed background is identical
```
EricThomson, line 356: background is messy b/c of CNMFE this is the major remaining hurdle I think.  But of course Eftychios and Co already foresaw this and implemented it with `compute_background()` :smile: 

So I think in general this would be very useful to have as its own thing.

EricThomson (commented):
In general seems like it will be useful especially for fpl stuff moving forward. I have mostly superficial concerns with naming and notational conventions. Only substantive concern is background computation stuff -- I think it would be good to have it in here rather than insert it as technical debt :computer:

caiman/source_extraction/cnmf/lazy_arrays.py
```diff
@@ -0,0 +1,444 @@
...
+    def __repr__(self):
+        r = super().__repr__()
+        return f"{r}" \
+               f"n_components: {self.n_components}"
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            raise TypeError(f"cannot compute equality against types that are not {self.__class__.__name__}")
+
+        if (self.spatial == other.spatial) and (self.temporal == other.temporal):
```
pgunn, line 350: Is there a reason not to do

`return self.spatial == other.spatial and self.temporal == other.temporal`

? Not insisting on this, just curious why it's spelled out in a bulky way here.

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: But it is outer product?

kushalkolar: Vector for spatial and vector for temporal are multiplied to product a matrix if I understand it correctly: https://en.wikipedia.org/wiki/Outer_product#Definition

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: It's easier to read if it's explicitly spelt out, requires less thinking while reading.

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: The key trick is `self.spatial @ self.temporal[:, indices]`, iiic, and this is basically `AC_`, or matrix multiplication on limited number of time points (columns) of C.  So I don't think it's using outer products explicitly. I guess it's outer product in the broad sense that all matrix multiplication is a sum of outer products :smile: , but it's basically just matrix multiplication limited to certain columns of the second matrix. Similarly for background `bf` it will be the same (except limited columns of bg fluctuations matrix `f`).

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: Hmmm- actually thinking about it a bit more.  Ignoring ssub_B for now assuming subsampling is 1 for the background fitting, the background in the CNMFE model is:

    background_model = b0 + W(Yr - AC - b0)

This is basically combining equations 5 and 8 from the CNMFE paper. 

Hmm maybe thought: all the things in the parens are do-able: b0 is easy (because it is multiplied by a vector of 1's of length T). AC is easy you already do that. Yr is already memmapped.  So basically subtraction operation inside parens, with some mult, and then matrix mult of W (which is huge) by that thing in the parens (W is like dims x dims). 

I think it can be done. And then we can get around the caveat in the code for compute background it throws a warning:

     logging.warning("Computing the full background has big memory requirements!")

:smile: 

It would take some more figuring out details with all this, and especially the `ssub_B` bits which I left out, but I think this should be doable with the framework you've built!

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: Isn't it outer product because you're doing vector x  vector = matrix? Andrea always referred to it as the outer product too.

kushalkolar: I generally thought:

vector.T * vector = single number, therefore inner product
matrix * matrix = matrix, two transformation so you just get another matrix
vector * vector.T = matrix, outer product

kushalkolar: maybe we can email Gibert Strang :joy:

caiman/source_extraction/cnmf/lazy_arrays.py
kushalkolar: > Yr is already memmapped

Always?? Is this a CNMFE specific thing?

> Once that operation is done, then matrix mult of W (which is a huge matrix). The nice thing is we wouldn't need to do all columns of the things inside the parens because lazy array.

Yes, `W` can have shape that is large, but still in the order of megabytes so it's not that bad. W is dense, the others are still sparse (right?) so it's should still be fast. 

> I think there are repercussions of this lazy array stuff that we just don't appreciate, tbh, that will be really cool as we work out the implications of the math in the models etc

Right now some of the most useful things are the projections, but there definitely must be other uses of lazy computing we can take advantage of. I wonder how much of dask's optimization engine is smart enough to catch all these things. For example, if it is able to figure out the standard deviation of the residuals without calculating the full thing that will be amazing.

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: But this is matrix x matrix, generally, right?

Here we have `A` (spatial component matrix in `self.spatial`) times some truncated `C` in `self.temporal[:, indices]` (indices picked based on how lazy you want to be). It's either matrix times vector (if you are picking just one index) or matrix time matrix (if you pick a block of indices). Not sure what I'm missing here it's just AC (but C is not all of C, because we're only picking certain columns).

Where we talk about outer product is when building up one component of the full neural model: a_i ⊗ c_i (one column from A and one row of C) and that gives you the single neuron movie.  But this is different, this is `AC_` (where I am using C_ to refer to the C matrix block limited to `indices` columns. Unless I'm missing something?

caiman/source_extraction/cnmf/lazy_arrays.py
EricThomson: I am not aware of `Yr` being used in a non-memmapped way in Caiman, but I may be wrong. We are pretty careful with that because it is huge.

PR #1174, opened 2023-09-20
Bokeh plotting 2->3 fix second (final?) round

EricThomson:
# Description

As far as I can tell, fixes the residual issues brought up in #1127 and previous PR #1129 . The problem was that images were inverted in ways I didn't notice, and this was percolating into almost every visualization function. I think it is now fixed in all tools (2d and 3d). 

# Type of change
[X] Bug fix (non-breaking change which fixes an issue)

# Has your PR been tested?
Yes against test and demotest, and I also tested with the following pipelines:

    demo_pipeline
    demo_pipeline_CNMFE
    demo_caiman_cnmf_3D
    demo_online_cnmfE
    demo_online_3d

These all seemed to work well!

@j-friedrich it would be helpful if you tested that this is indeed reasonable (esp in `demo_caiman_cnmf_3D.ipynb` which is where thigns broke before: I think `axis=2` is working well now too for the 3d component viewer with max projection, which is really nice, as that is the more intuitive projection).

### DIFF
--- a/caiman/source_extraction/cnmf/estimates.py
+++ b/caiman/source_extraction/cnmf/estimates.py
@@ -5,7 +5,7 @@ Created on Thu Jul 12 11:11:45 2018
 
 @author: epnevmatikakis
 """
-
+import bokeh
 import logging
 import matplotlib.pyplot as plt
 import numpy as np
@@ -249,7 +249,6 @@ class Estimates(object):
                 set of dictionary containing the various parameters
         """
         try:
-            import bokeh
             if 'csc_matrix' not in str(type(self.A)):
                 self.A = scipy.sparse.csc_matrix(self.A)
             if self.dims is None:
@@ -307,7 +306,7 @@ class Estimates(object):
                                use_cnn=params.quality['use_cnn'])
                 bokeh.plotting.show(bokeh.layouts.row(p1, p2))
         except:
-            print("Bokeh could not be loaded. Either it is not installed or you are not running within a notebook")
+            print("Error with bokeh plotter.")
             print("Using non-interactive plot as fallback")
             self.plot_contours(img=img, idx=idx, thr_method=thr_method,
                                thr=thr, params=params, cmap=cmap)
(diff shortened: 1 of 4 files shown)


caiman/source_extraction/cnmf/estimates.py
```diff
@@ -249,7 +249,6 @@ def plot_contours_nb(self, img=None, idx=None, thr_method='max',
                 set of dictionary containing the various parameters
         """
         try:
-            import bokeh
```
pgunn, line 252: Awesome to see imports moved out of conditionals

demos/notebooks/demo_caiman_cnmf_3D.ipynb
```diff
@@ -546,10 +543,7 @@
    "metadata": {},
    "source": [
     "### View the results\n",
-    "For a change we view the components as max-projections (frontal in the XY direction, sagittal in YZ direction and transverse in XZ), and we additionaly show the denoised trace\n",
-    "\n",
-    "**Note `nb_view_components_3d()` is buggy since a new Bokeh release, and we will work on a fix soon.     \n",
-    "If you know Bokeh and would like to help, please let us know!**"
+    "For a change we view the components as max-projections (frontal in the XY direction, sagittal in YZ direction and transverse in XZ), and we additionaly show the denoised trace"
```
pgunn, line 546: I know it was the original wording, but "For a change" is really weird phrasing; might be worth rephrasing it while it's being touched.

pgunn:
Generally LGTM

pgunn:
oasis.cpp really needs to be removed from this; if you don't mind removing it from the history of this branch that'd be ideal.

EricThomson:
> oasis.cpp really needs to be removed from this; if you don't mind removing it from the history of this branch that'd be ideal.

I keep trying new ways to eradicate that and this time I ended up committing it. I wonder if we add `*oasis.cpp` to `.gitignore` we'll have more success.

PR #1175, opened 2023-09-22
pyproject / PEP-517

pgunn:
This adds the needed additional files for PEP-517 compliant pip plumbing.

The practical impact of this is little:
* future-proofing our setup (in case the current tools eventually stop supporting old-style setup.py)
* For the very few people who build with environment.yml and then pip install, it is used and re-enables the `-e` flag
* It may make the (not encouraged but still sort-of-supported) pip install path used as a last resort and by colab a bit more efficient

In the longer-term we'll want to follow up on this by moving more of the fields from "dynamic" values defined in setup.py into residing just in the toml - I don't *ever* want to do that with the version value, but the other fields are good candidates for that). I'm not doing that now because I'd like to sit in the middle for awhile until all the tools and libraries are certain to look in at least both places (maybe that's already the case, I don't know and don't care to wager - moving slowly is safe).

Will need to keep an eye out for if the conda plumbing is impacted by this, although I don't expect it to be.

### DIFF
--- a/_buildhelper_cython.py
+++ b/_buildhelper_cython.py
@@ -14,11 +14,11 @@
 # small as oasis). No more `pip install -e .`
 
 from Cython.Build import cythonize
-from distutils.command.build_ext import build_ext
 import numpy as np
 import os
 import sys
 from setuptools import setup, find_packages
+from setuptools.command.build_py import build_py as _build_py
 import setuptools.extension
 
 class build_py(_build_py):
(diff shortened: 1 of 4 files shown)


_buildhelper_cython.py
```diff
@@ -0,0 +1,52 @@
...
+# cython parts of the build.
+#
+# This was ported from logic originally in the
+# old-stype setup.py, in order to comply with the
+# new PEP-517/PEP-518 build style.
+#
+# We intentionally do not support editable builds, as
+# they are confusing and usually the wrong thing for
+# projects that have a built component, even one as
+# small as oasis). No more `pip install -e .`
```
EricThomson, line 14: Is this bit about pip install -e ." accurate?

EricThomson:
Looks good, I've tested it on Windows in a few different configurations and I think this will be great moving forward in terms of flexibility -- thanks for doing this!

_buildhelper_cython.py
pgunn: I wrote that before our conversation where you said it's something you like, and after the conversation I planned to remove it after whatever code changes were needed to support it (not knowing that we'd get support without any further overriding of methods).

I'll remove the comment.

PR #1177, opened 2023-09-26
Final poking and prodding before release

EricThomson:
# Final PR before release [?]

WIP doing final tests for release 1.10 -- Caiman Electrolyte, or whatever we are calling it. 😆 

This will be a series of small commits to the demos (py files, hopefully none needed to the ipynb files). 

Will run tests when done.  This is a WIP.  

I'll post a comment once I've run through all the demos and have finished making commits.

### DIFF
--- a/demos/general/demo_caiman_basic.py
+++ b/demos/general/demo_caiman_basic.py
@@ -155,9 +155,7 @@ def main():
     cnm2.save(cnm2.mmap_file[:-4]+'hdf5')
 
     # %% play movie with results (original, reconstructed, amplified residual)
-    display_results = False
-    if display_results:
-        cnm2.estimates.play_movie(images, magnification=4)
+    cnm2.estimates.play_movie(images, magnification=4);
 
     # %% STOP CLUSTER and clean up log files
     cm.stop_server(dview=dview)


demos/general/demo_behavior.py
```diff
@@ -42,9 +45,9 @@
...
 #%%
 def main():
-    pass # For compatibility between running under Spyder and the CLI
+    pass # For compatibility between running under IDE and the CLI
 
-#%%
+    #%%
     pl.ion()
 
     fname = [u'demo_behavior.h5']
```
pgunn, line 53: With py3, the u marker on the string is no longer necessary (I know you didn't touch this, it's just in view.

We also could remove all the spyder breakpoints if we want, converting any that are comments into regular comments, or we could just leave them. Not that important to adjust either, really.

EricThomson: I considered this, but these code cells are now used in VS Code and even Pycharm, so I think they are useful. For people in CLI it doesn't really matter, but in the IDE they can be pretty helpful.

pgunn: Fair enough.

demos/general/demo_OnACID_mesoscope.py
```diff
@@ -34,16 +36,18 @@
 from caiman.source_extraction import cnmf as cnmf
 from caiman.utils.utils import download_demo
 
+%matplotlib qt
```
pgunn, line 39: We can't do this; it is invalid syntax outside of ipython

demos/general/demo_OnACID.py
```diff
@@ -18,6 +18,7 @@
 from caiman.source_extraction import cnmf as cnmf
 from caiman.paths import caiman_datadir
 
+%matplotlib qt #otherwise vis will not work
```
pgunn, line 21: We can't do this (see below)

demos/general/demo_OnACID_mesoscope.py
pgunn: Actually, we may be able to do this with eval(), conditionally on if we're in ipython

EricThomson: Good idea I'll give it a shot.

demos/general/demo_behavior.py
```diff
@@ -4,6 +4,9 @@
 """
 Created on Sun Aug  6 11:43:39 2017
 
+Warning: this demonstrates the use of caiman.behavior, which is
```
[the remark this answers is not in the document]
EricThomson: Yes, alas :smile: I'll roll back the warning. "Final poking and prodding branch" is not the time to unleash what is effectively a deprecation before we've officially discussed it.

demos/general/demo_OnACID.py
EricThomson: Fixed in `patched qt matplotlib ipython magic` commit below. Good call.

demos/notebooks/demo_multisession_registration.ipynb
```diff
@@ -118,13 +126,13 @@
...
     "\n",
     "As an exercise, we can collect the traces of all neurons that were registered in all sessions. We already gathered the indices of these neurons in the previous cell in `assignments_filtered`. Assuming that traces of each session are saved in their own `CNMF` object collected in a list, we can iterate through `assignments_filtered` and use these indices to find the re-registered neurons in every session.\n",
     "\n",
-    "Note: This notebook does not include the traces of the extracted neurons, only their spatial components. As such the loop below will produce an error. However, it demonstrates how to use the results of the registration to in your own analysis to extract the traces of the same neurons across different sessions."
+    "Note: This notebook does not include the traces of the extracted neurons, only their spatial components. As such the loop below will produce an error if you uncomment it. However, it demonstrates how to use the results of the registration to in your own analysis to extract the traces of the same neurons across different sessions."
```
pgunn, line 135: Would it be better to more deeply adjust the notebook so we don't need to have commented-out code? Or would that impose too much RAM or storage requirements on the notebook?

EricThomson: This cell always threw an error -- that code was just demonstrating how you might do something if you had data that we don't provide. Commenting it out is just to have the nb run without error.

pgunn: Should we make adjustments so the data is there to run the cell instead?

pgunn: (based on slack conversation, this is fine for now)

demos/general/demo_caiman_basic.py
```diff
@@ -137,27 +142,34 @@ def main():
...
     # cnm2.estimates.plot_contours(img=Cn)
+
     #%% save results
     cnm2.estimates.Cn = Cn
     cnm2.save(cnm2.mmap_file[:-4]+'hdf5')
 
-# %% play movie with results (original, reconstructed, amplified residual)
-    cnm2.estimates.play_movie(images, magnification=4)
+    # %% play movie with results (original, reconstructed, amplified residual)
+    display_results = False
```
pgunn, line 158: Why this switch? I'm not sure it's much of a demo if it just runs and exits and the user sees nothing.

demos/general/demo_pipeline.py
```diff
@@ -197,27 +198,29 @@ def main():
     cnm = cnmf.CNMF(n_processes, params=opts, dview=dview)
     cnm = cnm.fit(images)
 
-# %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE
+    # %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE (optional)
```
pgunn, line 201: At some point we'll need to clean up this phrasing more throughly (lowercase it too), but I'm not sure what "at once" is trying to say.

demos/general/demo_pipeline.py
```diff
@@ -269,7 +276,9 @@ def main():
         os.remove(log_file)
 
 # %%
-# This is to mask the differences between running this demo in Spyder
+# This is to mask the differences between running this demo in an IDE
 # versus from the CLI
 if __name__ == "__main__":
     main()
+
```
pgunn, line 283: ?

demos/general/demo_pipeline_NWB.py
```diff
@@ -195,35 +195,37 @@ def main():
...
+    # %% RUN CNMF ON PATCHES
     # First extract spatial and temporal components on patches and combine them
     # for this step deconvolution is turned off (p=0)
 
     opts.change_params({'p': 0})
     cnm = cnmf.CNMF(n_processes, params=opts, dview=dview)
     cnm = cnm.fit(images)
 
-# %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE
+    # %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE (optional)
```
pgunn, line 207: same as above

demos/general/demo_pipeline_NWB.py
```diff
@@ -271,18 +274,18 @@ def main():
...
 
 # %%
-# This is to mask the differences between running this demo in Spyder
+# This is to mask the differences between running this demo in IDE
 # versus from the CLI
 if __name__ == "__main__":
-    main()
\ No newline at end of file
+    main()
+# %%
```
pgunn, line 291: Why have a breakpoint at the end?

demos/general/demo_pipeline_cnmfE.py
```diff
@@ -194,17 +194,17 @@ def main():
     print(min_corr) # min correlation of peak (from correlation image)
     print(min_pnr)  # min peak to noise ratio
 
-# %% RUN CNMF ON PATCHES
+    # %% RUN CNMF ON PATCHES
     cnm = cnmf.CNMF(n_processes=n_processes, dview=dview, Ain=Ain, params=opts)
     cnm.fit(images)
 
-# %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE
+    # %% ALTERNATE WAY TO RUN THE PIPELINE AT ONCE (optional -- commented out)
```
pgunn, line 201: Same as above

demos/general/demo_pipeline_voltage_imaging.py
```diff
@@ -7,9 +7,11 @@
 Dataset courtesy of Karel Svoboda Lab (Janelia Research Campus).
 author: @caichangjia
 """
+#%%
```
pgunn, line 10: A breakpoint before the imports?

demos/general/demo_caiman_basic.py
EricThomson: This is the standard in all the notebooks where we suppress the opencv displays and people have to opt in. With this in particular it is run in demotests and I don't think we want opencv window popping up for tests (could be wrong we can change it back). There are other plots showing that use matplotlib.

demos/general/demo_pipeline_voltage_imaging.py
```diff
@@ -233,34 +234,34 @@ def main():
 
     opts.change_params(params_dict=opts_dict);          
 
-#%% TRACE DENOISING AND SPIKE DETECTION
+    #%% TRACE DENOISING AND SPIKE DETECTION
     vpy = VOLPY(n_processes=n_processes, dview=dview, params=opts)
-    vpy.fit(n_processes=n_processes, dview=dview)
+    vpy.fit(n_processes=n_processes, dview=dview);
```
pgunn, line 239: Is this semicolon just there for cell-based evaluation? Does it work that way? I thought that was just a jupyter thing

demos/notebooks/demo_motion_correction.ipynb
```diff
@@ -393,9 +394,9 @@
    "outputs": [],
    "source": [
     "# print crispness values\n",
-    "print('Crispness original: '+ str(int(crispness_orig)))\n",
-    "print('Crispness rigid: '+ str(int(crispness_rig)))\n",
-    "print('Crispness elastic: '+ str(int(crispness_els)))"
+    "print('Crispness original: ' + str(int(crispness_orig)))\n",
```
pgunn, line 397: f-strings might be more elegant

demos/general/demo_pipeline.py
EricThomson: It is explained a _little_ better in ipynb -- it runs motion correction, memory mapping, and cnmf in one method instead of breaking it up and running them separately. The py files are not a paragon of clarity. At some point we can loop around to actually improve them up my main goal here was to just make sure they work.

pgunn:
Generally looks good, left a few comments, if there are no substantial additions then we're ready to merge

demos/general/demo_pipeline.py
EricThomson: oops. I'll fix

demos/general/demo_caiman_basic.py
pgunn: It makes sense for notebooks, but for CLI demos we're probably expecting them not to actively edit the code so much; a different default in my view makes sense for them because it never makes sense for a CLI demo to quietly run and do nothing visible by default.

demos/general/demo_pipeline.py
pgunn: Yeah, it's just weird phrasing. "AT ONCE"

Maybe "all in one go" is a better phrasing?

demos/notebooks/demo_motion_correction.ipynb
EricThomson: Yes this is a pretty old notebook it's third on my list to renovate eventually, but for now I think let's just leave it.

caiman/base/movies.py
```diff
@@ -17,6 +17,7 @@
 from IPython.display import display, Image
 import ipywidgets as widgets
 import logging
+import matplotlib.pyplot as plt
```
pgunn, line 20: I think this new import is not needed? doesn't seem to be used anyhow

EricThomson: Good eye I was plotting something in an old debugging step.

PR #1181, opened 2023-10-04
main->dev before dev->main

pgunn:
We messed up the commit discipline, so we need to backmerge to dev


pgunn:
Resolving merge conflict with a guess as to what was intended

PR #1194, opened 2023-10-17
get_file_size move

pgunn:
This:
* moves get_file_size() from caiman/source_extraction/cnmf/utilities.py into base/movies.py, near load() and load_iter() for future refactoring. It also gets a partial type signature.
* Removes an unused do-not-use function that's been marked as being under development long after the people who might've been working on it left the project
* Also removes an entire file that was a junk drawer
* Does some import cleanups here and there


caiman/summary_images.py
```diff
@@ -14,8 +14,8 @@
 from typing import Any, Optional
 
 import caiman as cm
+import caiman.base.movies
```
EricThomson, line 17: This does make the code more readable.

caiman/utils/unused.py
EricThomson, line 1: It is a sad day when our repo no longer contains unused.py

EricThomson (commented):
Looks good to me!

PR #1202, opened 2023-10-19
remove relative imports, generally clean up imports, remove dead code

pgunn:
This gets rid of most relative imports in the codebase and removes some dead code

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -34,6 +34,7 @@ from zipfile import ZipFile
 import caiman as cm
 import caiman.base.movies
 import caiman.base.timeseries
+import caiman.base.traces
 import caiman.mmapping
 import caiman.summary_images
 import caiman.utils.visualization
@@ -1587,7 +1588,7 @@ def load(file_name: Union[str, list[str]],
 
         elif extension == '.mmap':
             filename = os.path.split(file_name)[-1]
-            Yr, dims, T = caiman.memmapping.load_memmap(
+            Yr, dims, T = caiman.mmapping.load_memmap(
                 os.path.join(                  # type: ignore # same dims typing issue as above
                     os.path.split(file_name)[0], filename))
             images = np.reshape(Yr.T, [T] + list(dims), order='F')
@@ -2284,7 +2285,7 @@ def get_file_size(file_name, var_name_hdf5='mov') -> tuple[tuple, Union[int, tup
                         dims[0], dims[1] = pims_movie.frame_shape[0:2]
             elif extension == '.mmap':
                 filename = os.path.split(file_name)[-1]
-                Yr, dims, T = caiman.memmapping.load_memmap(os.path.join(
+                Yr, dims, T = caiman.mmapping.load_memmap(os.path.join(
                         os.path.split(file_name)[0], filename))
             elif extension in ('.h5', '.hdf5', '.nwb'):
                 # FIXME this doesn't match the logic in load()
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -32,7 +32,6 @@ import warnings
 from zipfile import ZipFile
 
 import caiman as cm
-import caiman.base.movies
 import caiman.base.timeseries
 import caiman.base.traces
 import caiman.mmapping
@@ -2352,7 +2351,7 @@ def get_file_size(file_name, var_name_hdf5='mov') -> tuple[tuple, Union[int, tup
         else:
             raise Exception('File not found!')
     elif isinstance(file_name, tuple):
-        dims = caiman.base.movies.load(file_name[0], var_name_hdf5=var_name_hdf5).shape
+        dims = load(file_name[0], var_name_hdf5=var_name_hdf5).shape
         T = len(file_name)
 
     elif isinstance(file_name, list):
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -32,24 +32,18 @@ import warnings
 from zipfile import ZipFile
 
 import caiman as cm
-
-from . import timeseries
+import caiman.base.movies
+import caiman.base.timeseries
+import caiman.mmapping
+import caiman.summary_images
+import caiman.utils.visualization
 
 try:
     cv2.setNumThreads(0)
 except:
     pass
 
-from . import timeseries as ts
-from .traces import trace
-
-from ..mmapping import load_memmap
-from ..utils import visualization
-from .. import summary_images as si
-from ..motion_correction import apply_shift_online, motion_correct_online
-
-
-class movie(ts.timeseries):
+class movie(caiman.base.timeseries.timeseries):
     """
     Class representing a movie. This class subclasses timeseries,
     that in turn subclasses ndarray
@@ -885,7 +879,7 @@ class movie(ts.timeseries):
         T = self.shape[0]
         Cn = np.zeros(self.shape[1:])
         if T <= 3000:
-            Cn = si.local_correlations(np.array(self),
+            Cn = caiman.summary_images.local_correlations(np.array(self),
                                        eight_neighbours=eight_neighbours,
                                        swap_dim=swap_dim,
                                        order_mean=order_mean)
@@ -895,7 +889,7 @@ class movie(ts.timeseries):
             for jj, mv in enumerate(range(n_chunks - 1)):
                 logging.debug('number of chunks:' + str(jj) + ' frames: ' +
                               str([mv * frames_per_chunk, (mv + 1) * frames_per_chunk]))
-                rho = si.local_correlations(np.array(self[mv * frames_per_chunk:(mv + 1) * frames_per_chunk]),
+                rho = caiman.summary_images.local_correlations(np.array(self[mv * frames_per_chunk:(mv + 1) * frames_per_chunk]),
                                             eight_neighbours=eight_neighbours,
                                             swap_dim=swap_dim,
                                             order_mean=order_mean)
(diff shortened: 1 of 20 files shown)


caiman/source_extraction/cnmf/initialization.py
```diff
@@ -283,11 +278,6 @@ def initialize_components(Y, K=30, gSig=[5, 5], gSiz=None, ssub=1, tsub=1, nIter
 
     """
     method = method_init
-    if method == 'local_nmf':
```
pgunn, line 286: Note that I'm removing this because it never worked (in fact could not have worked because below it depends on imports that didn't work)

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -27,22 +26,22 @@
...
+from caiman.motion_correction import MotionCorrect
 import caiman.paths
+from caiman.source_extraction.cnmf.estimates import Estimates
+from caiman.source_extraction.cnmf.initialization import initialize_components, compute_W
+from caiman.source_extraction.cnmf.map_reduce import run_CNMF_patches
+from caiman.source_extraction.cnmf.merging import merge_components
+from caiman.source_extraction.cnmf.params import CNMFParams
+from caiman.source_extraction.cnmf.pre_processing import preprocess_data
+from caiman.source_extraction.cnmf.spatial import update_spatial_components
+from caiman.source_extraction.cnmf.temporal import update_temporal_components, constrained_foopsi_parallel
```
EricThomson, line 41: These I like. I won't have to read through a bunch of dots to get to the method/function name in the code. 

Some of the other ones (e.g., in `movies.py`) I worry the readability suffers.

EricThomson: I'm having trouble coming up with a general rule for this. Sometimes I want a dot, maybe even two. Often I want none. It is probably impossible to come up with a general algorithm based on length of variables or number of dots, because it has to do with what seems "natural" after spending a lot of time reading the code. These ones seems natural. (Some) of the ones in `movies.py` seem less natural and a bit more like they were done by a refactoring algorithm (not saying they were but you get my point I hope). 

I'll take a bit of time in `movies.py` to illustrate what I'm thinking.

caiman/base/movies.py
```diff
@@ -1014,7 +1008,7 @@ def resize(self, fx=1, fy=1, fz=1, interpolation=cv2.INTER_AREA):
                 if len(new_m) == 0:
                     new_m = m_tmp
                 else:
-                    new_m = timeseries.concatenate([new_m, m_tmp], axis=0)
+                    new_m = caiman.base.timeseries.concatenate([new_m, m_tmp], axis=0)
```
EricThomson, line 1011: For instance, here I think I'd want `timeseries.concatenate()`

caiman/base/movies.py
```diff
@@ -885,7 +879,7 @@ def local_correlations(self,
         T = self.shape[0]
         Cn = np.zeros(self.shape[1:])
         if T <= 3000:
-            Cn = si.local_correlations(np.array(self),
+            Cn = caiman.summary_images.local_correlations(np.array(self),
```
EricThomson, line 882: would prefer just `local_correlations()` or at most `summary_images.local_correlations()`

caiman/base/movies.py
```diff
@@ -32,24 +32,18 @@
 from zipfile import ZipFile
 
 import caiman as cm
-
-from . import timeseries
+import caiman.base.timeseries
+import caiman.base.traces
```
EricThomson, line 36: would import as `traces`

caiman/base/movies.py
```diff
@@ -1593,7 +1587,7 @@ def load(file_name: Union[str, list[str]],
 
         elif extension == '.mmap':
             filename = os.path.split(file_name)[-1]
-            Yr, dims, T = load_memmap(
+            Yr, dims, T = caiman.mmapping.load_memmap(
```
EricThomson, line 1590: For reasons I can't explain very well, I like this expanded version of this as it says this is a caiman memory mapping mechanism we are getting into (people might think it is numpy or something or get confused). There is no hard fact of the matter here about the right cutoff I'm just throwing my intuitions here and doing my best to explain them 😬

caiman/base/movies.py
```diff
@@ -1693,7 +1687,7 @@ def load_movie_chain(file_list: list[str],
             m = m[:, top:h - bottom, left:w - right, z_top:d - z_bottom]
 
         mov.append(m)
-    return ts.concatenate(mov, axis=0)
+    return caiman.base.timeseries.concatenate(mov, axis=0)
```
EricThomson, line 1690: I think in general my feeling is `timeseries.concatenate()` would be better.

caiman/base/movies.py
```diff
@@ -2537,7 +2530,7 @@ def animate(i):
         anim = matplotlib.animation.FuncAnimation(fig, animate, frames=frames, interval=1, blit=True)
 
         # call our new function to display the animation
-        return visualization.display_animation(anim, fps=fr)
+        return caiman.utils.visualization.display_animation(anim, fps=fr)
```
EricThomson, line 2533: Would prefer either `display_animation()` or at worst `visualization.display_animation()`.

caiman/base/rois.py
```diff
@@ -21,7 +21,7 @@
 from typing import Any, Optional
 import zipfile
 
-from ..motion_correction import tile_and_correct
+import caiman.motion_correction
```
EricThomson, line 24: Prefer `as motion_correction` here

caiman/base/traces.py
```diff
@@ -1,25 +1,19 @@
...
-#%%
+
 import cv2
 import logging
 import numpy as np
 import pylab as pl
 pl.ion()
-from . import timeseries as ts
+
+import caiman.base.timeseries
```
EricThomson, line 9: in general with `caiman.base` I would prefer to import the final object. E.g., just import `timeseries` because `caiman.base.timeseries.timeseries` doesn't help me read the code.

caiman/components_evaluation.py
```diff
@@ -13,9 +13,10 @@
 from typing import Any, Union
 import warnings
 
+import caiman
 from caiman.paths import caiman_datadir
-from .utils.stats import mode_robust, mode_robust_fast
-from .utils.utils import load_graph
+import caiman.utils.stats
```
EricThomson, line 18: Generally I think `caiman.utils` I would just import `stats` so instead of having to read `caiman.utils.stats` later I just have to read `stats.mode_robust` later.

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1,25 +1,28 @@
...
-from .merging import merge_iteration, merge_components
-from ...components_evaluation import (
-        evaluate_components_CNN, estimate_components_quality_auto,
-        select_components_from_metrics, compute_eccentricity)
-from ...base.rois import (
-        detect_duplicates_and_subsets, nf_match_neurons_in_binary_masks,
-        nf_masks_to_neurof_dict)
-from .initialization import downscale
-
+import caiman.base.rois
```
EricThomson, line 19: I think I am coming to a one dot limit in what I want to read later in the code.

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1,25 +1,28 @@
...
+import caiman.components_evaluation
+import caiman.source_extraction.cnmf.initialization
```
EricThomson, line 21: I would suggest importing `initialization` so later I can see `initialization.downscale`

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -725,7 +727,7 @@ def compute_background(self, Yr):
             if ssub_B == 1:
                 return self.b0[:, None] + self.W.dot(Yr - self.A.dot(self.C) - self.b0[:, None])
             else:
-                ds_mat = decimation_matrix(self.dims, ssub_B)
+                ds_mat = caiman.source_extraction.cnmf.utilities.decimation_matrix(self.dims, ssub_B)
```
EricThomson, line 730: This is not good readin' :smile:

caiman/source_extraction/cnmf/estimates.py
```diff
@@ -1491,7 +1496,7 @@ def masks_2_neurofinder(self, dataset_name):
             raise Exception(
                 'You need to compute thresholded components before calling this method: use the threshold_components method')
         bin_masks = self.A_thr.reshape([self.dims[0], self.dims[1], -1], order='F').transpose([2, 0, 1])
-        return nf_masks_to_neurof_dict(bin_masks, dataset_name)
+        return caiman.base.rois.nf_masks_to_neurof_dict(bin_masks, dataset_name)
```
EricThomson, line 1499: I would suggest import `rois` alone.

caiman/source_extraction/cnmf/params.py
```diff
@@ -6,13 +6,12 @@
...
-from scipy.ndimage import generate_binary_structure, iterate_structure
+import scipy.ndimage
 
 import caiman.utils.utils
 import caiman.base.movies
-from ...paths import caiman_datadir
-from .utilities import dict_compare
-
+import caiman.paths
+import caiman.source_extraction.cnmf.utilities
```
EricThomson, line 14: Not going to keep going through the remaining files, but I just want to say I really like the overall gist of what this PR does with making imports absolute. It makes things **way** more clean and clear! 🚀 

 Where I am quibbling is just in some of the details about where some of the specific cutoffs happen in the dots :smile: When reading the code, it can be a bit cumbersome/disruptive to wade through. Basically if there are things like `caiman.base.utils.visualize.cnmf.view.outlier` coming before a function, then we should trim it back. I would just focus on that but I think this is a really god improvement.

EricThomson (commented):
I really like this -- getting rid of relative imports makes things way more clean and readable. As I mentioned in my last comment was that I'd only quibble about the cutoff of some of the imports.

EricThomson:
One thing I've never thought about: why are relative imports considered bad in general, but ok in `__init__.py` files?  Is this something we are ok with?

pgunn:
I normally don't look at those because they're not really code, but they should be changed too. They're just really easy to forget about.

pgunn:
I let this sit for too long and am going to redo this diff against the current dev branch, taking into account some of the feedback and rolling back a few things (and adjusting others). Will hold this PR open until the diff that comes from it lands, to act as a reference for the redo.

pgunn:
Closing this because it became too stale to easily fix; succeeded by #1277

Issue #1212, opened 2023-10-25
Tracking: setup_cluster and None


pgunn:
This is a tracking issue to explore reworking setup_cluster() to allow for explicitly saying "I don't want a cluster just run it single-threaded" without external tricks, having the returned value store that and having other functions that work on a dview understand that.

It'd be a fairly involved change, but would make handling "how things run" more regular.

pgunn:
This may be less work than I thought; what probably would be best would be to rework the code behind the single_thread flag to instead be a different named backend, deprecating single_thread. That'd be way clearer.

pgunn:
We could still modify it not to return None, but we might defer such refactoring until later potential refactoring that might put dviews and temp/output paths under some common object.

PR #1214, opened 2023-10-25
cluster.py:setup_cluster() - cleanups, make 'single' backend

pgunn:
See tracking issue #1212

### DIFF
--- a/caiman/cluster.py
+++ b/caiman/cluster.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 
 """ functions related to the creation and management of the "cluster",
 meaning the framework for distributed computation.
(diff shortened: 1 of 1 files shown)


pgunn:
This also reorders the handlers in order of use; leading with the (rarely-used, probably-does-not-work) slurm handler didn't make sense.

It makes a token attempt to allow for the slurm backend to work by making the slurm script an environment variable rather than the old hardcoded path in Andrea's home directory (haha)

caiman/cluster.py
```diff
@@ -337,26 +336,31 @@ def stop_server(ipcluster: str = 'ipcluster', pdir: str = None, profile: str = N
...
-            'multiprocessing' [alias 'local'], 'ipyparallel', and 'SLURM'
-            ipyparallel and SLURM backends try to restart if cluster running.
-            backend='multiprocessing' raises an exception if a cluster is running.
-        n_processes: int
+        backend
+            One of:
+                'multiprocessing' - Use multiprocessing library
+                'ipyparallel' - Use ipyparallel instead (better on Windows?)
+                'single' - Don't be parallel (good for debugging, slow)
+                'SLURM' - Try to use SLURM batch system (untested, involved).
```
EricThomson, line 352: I like this!

EricThomson (approved):
Generally: good change I like the mod it will make things more understandable and clean moving forward.

PR #1217, opened 2023-10-27
Document file formats people can bring to caiman

pgunn:
This is meant to be a guide on what file formats Caiman can work with, and to some extent how to prep data for use with Caiman.

It needs work (Eric, Kushal, please join me in working on the branch to get this thing up to snuff). Some sections explicitly note that they're incomplete. We should add stuff on converters.

We should have documented this ages ago. Still, let's not let that be an excuse not to document it now.


EricThomson:
Great idea, will probably not be able to do much with this before workshop next week, but after that I will have a surplus of time!

docs/file_formats.md
```diff
@@ -0,0 +1,64 @@
...
+- get\_file\_size()
+
+If you want to implement your own handlers with a fork of our codebase, or submit a PR to us, start there. Caiman converts files from these formats into its own internal formats during motion correction.
+
+Notes for specific formats follow:
+### Matlab (\*.mat)
+Caiman may be able to handle Matlab's mat files, although it is not a desirable format. We rely on the scipy.io.matlab functions to handle files of this type, except for certain versions of the format which are actually hdf5 files
+
+### TIFF (\*.tiff, \*.tif, \*.btf)
+Caiman can handle many TIFF files, but TIFF is a very flexible file format and we do not handle every variant. Because calcium imaging data has a time element, there are many ways that files might saved in this format (series, pages, some other things). We don't anticipate every variant.
```
kushalkolar, line 19: Perhaps say the best thing to do is just try, and if it doesn't work then use @EricThomson tiffit to convert it? https://github.com/EricThomson/tiffit

docs/file_formats.md
```diff
@@ -0,0 +1,64 @@
...
+## Functions
+load() supports a broader range of file formats than load\_iter().
+(TODO fill in this section)
+
+On bit-depth for the formats
+(TODO fill in this section)
+
+## Conversion tools
+ImageJ
+(tiff file converters)
```
kushalkolar, line 51: https://github.com/EricThomson/tiffit

docs/file_formats.md
```diff
@@ -0,0 +1,64 @@
...
+
```
kushalkolar, line 52: What about adding an example of how to convert something that can be loaded as a numpy array into a caiman-compatible format? Might be useful for beginners.

docs/file_formats.md
pgunn: Let's do that, but let's find/list a few more as well. If we add tools but only things we write, it looks a little weird.

EricThomson: The problem is imagej produces things that tend to not work. This is the main reason I made tiffit -- to fix movies that people were saving with imagej and were not working with Caiman (tiffit is a lightweight wrapper for tifffile and tifftools). 

The other tools we could list include tifftools and tifffile (also ffmpeg, pillow, opencv, and pims are really good even though not tiff specific they are really good so maybe just mention them somewhere). They are much higher-powered tools (first two are not tiff specific).  There isn't a really good tool out there for tiff it is a really wild heterogeneous format (`thousands of incompatible file formats`).

pgunn: I filled it in more - I don't want to recommend API-only "tools", as if someone is comfortable with coding, they don't really that kind of help, and there are too many APIs that can do this to list.

docs/file_formats.md
```diff
@@ -39,16 +39,11 @@ Support for this format was removed in an earlier version of Caiman.
...
-
-On bit-depth for the formats
-(TODO fill in this section)
-
 ## Conversion tools
-ImageJ
-(tiff file converters)
+- ImageJ - A powerful (but high learning curve) tool that can visualise and convert a number of formats
+- TiffIt - https://github.com/EricThomson/tiffit - Can convert incompatible tiff files into compatible tiff files
+- GIMP - A general-purpose image editor that can (clumsily) work with video files. Scriptable with a number of useful filters
```
kushalkolar, line 45: gimp is very impractical for video files from my experince

pgunn:
I'm thinking of landing this soon; we can do more work on it later.

PR #1228, opened 2023-11-15
add test workflow

bendichter:
# Description

Add continuous integration. This is running and working here: https://github.com/bendichter/CaImAn/pull/2

Fixes # (issue)

# Type of change

# Branching
- All PRs should be made against the **dev** branch. The master branch is not often merged back to dev.

# Has your PR been tested?

```python caimanmanager.py test```

and

```python caimanmanager.py demotest```

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -1,0 +1,41 @@
+name: Linux Conda test
+
+on:
+  pull_request:
+    branches:
+      - master
+      - dev
+  push:
+    branches:
+      - master
+
+jobs:
+  build-linux-conda:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 5
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python '3.11'
+        uses: actions/setup-python@v3
+        with:
+          python-version: '3.11'
+      - name: Add conda to system path
+        run: |
+          # $CONDA is an environment variable pointing to the root of the miniconda directory
+          echo $CONDA/bin >> $GITHUB_PATH
+
+      - name: Install mamba
+        run: |
+          conda install -c conda-forge mamba
+          conda clean -a
+      - name: install caiman
+        run: |
+          mamba env update -n base --file environment.yml
+          pip install -e .
+          caimanmanager install --inplace
+
+      - name: Test
+        run: |
+          caimanmanager test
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by EricThomson** [comment]
This sounds great. This kind of resuscitation (we turned off CI a couple of years ago I think) is something on my internal roadmap that I've been putting off until next year, so I appreciate this work it's pushing me to learn more about actions etc.   :rocket:

For the tests being run, would we also want to include `demotest`, or is that too heavy? What do you think @pgunn 

Also before a merge we'd need to change `master` to `main`

**Comment by pgunn** [comment]
demotest is probably too involved (and adds troublesome dependencies), so let's skip that. Otherwise, this generally looks good; I'm going to need to really be sure this won't cut into any kind of limits we have before we accept (if it does, that's a showstopper), but I'll try to find that out soon.

**Comment by bendichter** [comment]
Let's commit this minimal CI and build upon it as needed. I also think we should test across platform and across supported Python versions, but I just wanted to get this in first so we have a place to build from.

File: .github/workflows/run_tests.yml [hunk 1]
```diff
@@ -0,0 +1,41 @@
+name: Linux Conda test
+
+on:
+  pull_request:
+    branches:
+      - master
+      - dev
+  push:
+    branches:
+      - master
```
bendichter, line 10:
```suggestion
      - main
      - dev
  push:
    branches:
      - main
```

pgunn:
Looks like it's not just a forum that says that GA usage is free for public repos; it's documented as such here:

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

Given that, I have no reservations with use of GA for CI.

pgunn:
If it's not too much effort, we should have this make an environment called "caiman" and activate it, rather than update the base environment except to install mamba into it (which is fragile; a caiman install could easily break the base environment). This would also lessen concerns of accidentally changing the python version in the base env or something like that.

bendichter:
@pgunn I was having issue with that earlier but maybe I can get it to work now

pgunn:
I don't see where, in the changed file, it specifies that it wants a conda version of python (or where it installs conda), and available documentation seems to suggest that that step installs python from pip.

How does this thing work?

bendichter:
@pgunn here's a cleaner version with explicit conda installation and environment management

pgunn:
Looks good; is this ready for merge?

bendichter:
Yes, this is ready to merge.

PR #1230, opened 2023-11-16
Multi ci

bendichter:
# Description

Expands upon CI to test Python 3.9 through 3.11.


bendichter:
@pgunn an improvement that tests 3 different versions of Python in parallel

Issue #1213, opened 2023-10-25
Tracking: slurm cleanup


pgunn:
The code for the slurm backend to setup_cluster() is embarrassingly bad, with hardcoded paths. It also probably doesn't work (the codebase has a lot of explicit mention of python2.x).

We should either remove that code or clean it up. Urgently.

pgunn:
Made a small (untested, but might work) effort towards fixing this in #1214

pgunn:
The SLURM/ directory still needs a lot of attention, if we think it's worth even fixing this code. I'm pretty sure some of those imports won't work anymore. Even after fixing them, we'd need to figure out and document a lot of this stuff to give people a fighting chance.

EricThomson:
This would be really helpful IMO. There are people I've talked to offline, at workshops, who roll their own solution. They tend to be power users and not ask for help and sort of just figure it out on their own (they often have HPC people helping). But there are also probably a lot of people who would appreciate a nudge in the right direction with a little example shell script or two.

pgunn:
setup_cluster() is adjusted (since 1214 landed) and now could at least in theory work.

Next steps to satisfy this tracking issue:

1) Cleanup SLURM/
2) Test all of this
3) Document all of this

PR #1233, opened 2023-11-17
Remove slurm integration

pgunn:
Based on:
A) Looking over the code carefully and determining that it hasn't been tested since the Python 2.7 days and that it
B) Doesn't work with the modern codebase
C) Is undocumented
D) In the last released version of Caiman had dependencies on andrea's home directory hardcoded
E) Even if it were fully fixed would almost never provide a performance benefit given current levels of parallelism
F) And is really ugly/weird

this diff removes the old slurm integration from the code.

To be clear: some of our users run Caiman under slurm. This removal does not impact them because nobody we could find was using the slurm integration (given how broken it was and specific to our local cluster, it's unclear how they might've gotten it working). They were just allocating a node under slurm and running caiman on it, which is the same as running any other software on a node. Caiman doesn't notice SLURM by default and doesn't mind its specifics.


pgunn:
Resolves #1213

PR #1250, opened 2024-01-25
fix estimates.play_movie so range var works

### LINKED ISSUE #1164: A bug in play_movie function of estimates.py
Hi, in the current implementation of play_movie function, line 667, imgs[frame_range] is sliced the second time after it's done initially in the if block in line 620. This would gave error if one tries to play or save a portion of the movie. Thanks.

EricThomson:
# Description
Fixes #1164 

# Type of change

Reproduced suspected bug, and fix was simple and worked and tested in demo.


EricThomson:
Oops I left some print statements in there while debugging I'll remove. 😄

Issue #1262, opened 2024-02-01
Should block ability to stumble into infinite while loop in percentile calculation for dff


EricThomson:
When calculating dff there is a percentile calculation that has a try/except block in a while loop. If the exception is triggered every run, the while loop never breaks.  😱  I ran into this today. We should fix this design flaw

https://github.com/flatironinstitute/CaImAn/blob/344a7e5651736d99f77c0be90928d517fda0b9e9/caiman/utils/stats.py#L184

EricThomson:
Yikes this is a consequence of functions being deprecated in scipy. Will temporarily lock to < 1.12 for now and am adding to roadmap to fix.

kushalkolar:
Oh my, I cannot comprehend the amount of :spaghetti: in that function. This is also an example of why bare `except` is generally a bad idea. From a first glance this could be remedied by adding a `finally` block and putting a counter inside the `except` to force exiting the loop if more than `n` exceptions are raised. But given how deeply nested that function is, I think it can just be designed better but I don't have time for that.

EricThomson:
Kushal -- I started with a counter, but I ended up unhappy with this approach as it leaves the smelly code in there. 😄 As you said, it can be designed better. The problem in this case is that it depends (after you go down the stack enough) on a deprecated function, so it will always hit the counter limit. I need to do a real sweep of Caiman for soon-to-be deprecated functions! There are lots of them.

EricThomson:
Fixed with #1288

PR #1263, opened 2024-02-01
Minor tweaks in notebooks for next 1.10 patch

EricThomson:
Expected to go out in next patch (with image link fix #1261 ). Improved formatting of estimates description. Improved discussion of $\Delta F/F$.  Added example plot of DFF trace. 

Still need to look over a couple of things wrt DFF, e.g., would like to fix infinite loop (#1262 ).


EricThomson:
I should have mentioned how I "fixed" the dff infinite loop bug: avoid that section of code by setting `flag_auto`  to `False`. IOW technical debt...incurred. Need to fix the bug still.

PR #1270, opened 2024-02-09
Notebooks:images live in images/ subdir next to notebooks

pgunn:
Based on mostly in-person discussion, this reverts the recent change that uses github links to host images for the jupyter notebooks (for a few reasons - offline access, possibly problematic firewalls (including China's firewalls which occasionally block US/Europe hosted sites), Github maybe someday blocking third-party referrers for embedded images, avoiding future updates to images changing content in old software releases, and avoiding commitment to keeping those files exactly there to avoid breaking notebooks in old software), moving back to using the images that are installed with caimanmanager.

It also changes the docs that told users to cd into caiman_data/demos/notebooks before launching Jupyter, which leads to problems.

At the heart of this is a (security?) restriction Jupyter has for its HTML rendering - it refuses to serve images from locations where it would need to step backwards through the filesystem to find them (python is allowed to go anywhere, HTML path resolution is not - Jupyter doesn't want to potentially act as a webserver for your entire hard drive).

If in the future users see rendering issues like this we'll need to keep in mind that they may have launched jupyter from the wrong place.

Will leave this PR in place for a few days to allow for any discussion needed.


pgunn:
Based on further yet discussion, implementing an idea to put images used in notebooks in an images/ directory that lives right next to the notebooks; this is an even better solution because it separates those images from those actually used in docs (making the difference self-documenting) and gets rid of any sensitivity to where jupyter is launched because if the notebook is within scope of what jupyter can see, the images will be too. I need to test caimanmanager.py to be sure it will actually install these.

PR #1279, opened 2024-02-16
Fix some sphinx issues

pgunn:
This PR is an attempt to fix some of the sphinx rendering issues - namely the issues with the core functions having sections but no functions/methods showing up.

It does NOT fix all the issues we're aware of, particularly the ReadTheDocs broken links for non-function-reference stuff, but those will take some thought and a redesign.

At least with the local rendering, things look good. This can be reproduced with a source tree by:
1) activating whatever environment is used (I use `caibuild` for my compiled caiman environment)
2) doing `mamba install -c conda-forge sphinx numpydoc`
3) cd docs
4) `sphinx-build -M html source html`
5) Point your browser at the docs generated under html/html

For iterating over errors while changing and fixing things, I replace step 4 with:

`pushd .. ; pip install . ; popd ; rm -rf html ; clear ; sphinx-build -M html source html 2>&1 | less`

and look for the word ERROR (and at any warnings you see, if you like).

The bulk of this is fixing fiddly comment formatting; sphinx has very particular expectations on how docstrings must be formatted for functions for them to be turned into docs, and I only somewhat understand the rules, but this gets rid of all errors I see when running locally and the functions turn up locally too.

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -2203,23 +2203,24 @@ def load_iter(file_name: Union[str, list[str]], subindices=None, var_name_hdf5: 
             logging.error(f"File request:[{file_name}] not found!")
             raise Exception('File not found!')
 
-def get_file_size(file_name, var_name_hdf5='mov') -> tuple[tuple, Union[int, tuple]]:
-    """ Computes the dimensions of a file or a list of files without loading
+def get_file_size(file_name, var_name_hdf5:str='mov') -> tuple[tuple, Union[int, tuple]]:
+    """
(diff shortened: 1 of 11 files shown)


pgunn:
This diff changes no real functional code, so the chances of breaking something are pretty much nil. It's also what I suspect is both unobjectionable and uninteresting to anybody; will probably land this later today or tomorrow unless anyone objects; if it looks good on RTD we may do a release just to get this into the main branch there (will be a very uninteresting release, but that's okay).

PR #1286, opened 2024-02-20
Sphinx/github docs tweaks to fix broken links

### LINKED ISSUE #1278: Docs issue: paths/links differ between Sphinx/RTD and local/Github rendering, leading to broken links
Please fill in the following for any issues

Your setup:
 1. Operating System (Linux, MacOS, Windows): MacOS
 2. Hardware type (x86, ARM..) and RAM: x86, 16GB
 3. Python Version (e.g. 3.9): 3.11.7
 4. Caiman version (e.g. 1.9.12): 1.10.2
 5. Which demo exhibits the problem (if applicable):
 6. How you installed Caiman (pure conda, conda + compile, colab, ..): conda
 7. Details: 

Wanted to let you know that the link in the documentation for setting up GPU use is broken.
https://caiman.readthedocs.io/en/latest/README-GPU.md results in a 404 not found error:

404 Not Found

You are browsing the documentation of caiman. The documentation page you are looking for was not found.

Documentation changes over time and pages are moved around. You can try to navigate to [the index page](https://caiman.readthedocs.io/) of the project and use its navigation, or search for a similar page.

pgunn:
Attempts to fix #1278, also makes more docs available in sphinx/RTD and fixes some formatting


pgunn:
I am not entirely happy with how some of this works; I wasn't able to get any of the link-rewriting options to work, so I just revised the docs to not have those kinds of links. It solves the problem (and having more docs available in the doctree and fixing some of the unintended display issues in the doctree helps too).

There's more to fix here (tomorrow) because some of the docs I moved in don't render correctly. I might also see if there are other ways to rewrite URLs to make the cross-system each-link-with-local-semantics ideal happen, but we can actually do without that (because usually those links are actually bad content for things that make reading the docs feel scatterbrained); what's bad is the broken links.

pgunn:
Everything parses correctly now, even if some of the solutions here are not ideal. If we were to land this we'd fix many (all?) broken links on RTD and things would look better. I will still see tomorrow if there's a way to have consistent inner links within both distinct systems, but if we really needed to get this out the door this would at least be better than the way things are now.

pgunn:
(documenting for later that I tried recommonmark and MyST; neither seamlessly rewrote things and it was unclear how to add configuration requesting it do so)

EricThomson:
Looks good to me when I build locally and poke at things. It's hard to actually test locally that's part of the root cause of the problem, but seems good to go. There is a logo on the Overview page that's the only place I saw one in docs. We should consider whether we want to switch in new logo by keeping name same or changing name (if you know what I mean).

Issue #1283, opened 2024-02-19
Brittle Handling of failed `utils/stats.kde(...)` calls in `utils/stats.df_percentile`


CorbanSwain:
https://github.com/flatironinstitute/CaImAn/blob/c873b036135a4c2f441c78449d42eb887d356b7e/caiman/utils/stats.py#L183-L192

Given the broad catching of all exceptions; unexpected errors in `kde(...)` (i.e. any error not resolved by lengthening the input list) can lead to an infinite loop and exponential memory usage. I believe this was hanging Python when I encountered issue #1281 .

Example of my hotfix (but not necessarily to be duplicated) ... I think that the best solution would be to add the explicit exceptions intended to be caught when `kde()` fails after the `except` keyword.

```python
def df_percentils(...):
   # ...
   if axis is not None:
      # ...
   else:
        RETRY_LIMIT = 16
        retry_count = 0
        err = True
        while err:
            try:
                bandwidth, mesh, density, cdf = kde(inputData)
                err = False
            except AttributeError:
                raise
            except:
                retry_count += 1
                if retry_count > RETRY_LIMIT:
                    raise

                logging.warning('Percentile computation failed. Duplicating ' + 'and trying again.')
                if not isinstance(inputData, list):
                    inputData = inputData.tolist()
                inputData += inputData
```

pgunn:
This will take some thought; your approach seems reasonable, but I wonder if we might be able to detect how things are going wrong without retrying. If not, we might end up adopting roughly this solution.

kushalkolar:
This issue is also discussed here: https://github.com/flatironinstitute/CaImAn/issues/1262

EricThomson:
Yes duplicate. I'm working on this should fix this week.

pgunn:
Closing this as a dup

Issue #1274, opened 2024-02-13
'cnmf_refit.estimates.detrend_df_f' in demo_pipeline.ipynb requires tons of memory?


brian-arnold:
Please fill in the following for any issues

Your setup:
Repeatable on both my university's computing cluster and my Mac laptop

 1. Operating System (Linux, MacOS, Windows):
- Linux and MacOS (14.2)

 2. Hardware type (x86, ARM..) and RAM:
- Hardware: x86 and ARM 
- RAM: 150GB+ on computing cluster, 96GB on Mac

 3. Python Version (e.g. 3.9):
 - 3.11.7

4. Caiman version (e.g. 1.9.12):
- 1.10.2 

5. Which demo exhibits the problem (if applicable):
 - demo_pipeline.ipynb, ran cell by cell without making any changes, using the data provided by the tutorial: download_demo('Sue_2x_3000_40_-46.tif')

6. How you installed Caiman (pure conda, conda + compile, colab, ..):
- conda, but for my MAC I had to additionally 'pip install tensorflow-macos' to get the following command to run:
cnmf_refit.estimates.evaluate_components

7. Details: 

Thank you for such a fabulous set of well-documented tools!

I'm running demo_pipeline.ipynb, using data provided by the demo, and all but one cell in the notebook works like a charm, both when I run the jupyter notebook (via VS Code) interactively on our computing cluster with lots of memory and my Mac, which also has a decent amount of memory. 

However, every time I reach the cell containing the function that computes delta_f/f (using cnmf_refit.estimates.detrend_df_f), the kernel crashes. I suspect it's a memory issue because when I monitor memory consumption on our computing cluster using 'top', over the course of ~15 min memory usage gradually increases to ~200GB before it hits 0 and the kernel crashes. Likewise, on my Mac with 96GB of memory, after 15 min it says the application ran out of memory.

Are there any options or parameters I can change to help this? Or is the only solution to just get more memory? We have some large datasets we want to analyze, so I worry that if I'm hitting memory issues with the demo data that I will certainly encounter this issue on the data we're interested in.

Thanks for any help or advice!
Brian

EricThomson:
Yikes that's not good. This is a _relatively_ simple calculation that shouldn't require that much memory. Let me try to reproduce soon: we've started to adjust things around that method recently (see #1260 ). I should be able to get to it tomorrow.

kushalkolar:
What if you give `use_residuals=False` ?

[...]

PR #1288, opened 2024-02-21
Handle problem with dff auto-percentile calculation

EricThomson:
# Description
When `flag_auto` was set to `True` in dff calculation, there was a significant chance that things could end up in an infinite while loop. 

In the kernel density estimator, an exception which was rarely thrown was exposed recently with some scipy deprecations, so this has started happening a lot lately. It was handled previously with printing 'oops' and returning `None`. This PR fixes that, stops the while loop from recurring indefinitely. The basic logic wasn't terrible it's fine to use `try/except` blocks as control elements in Python, it's just a bad idea to let them go on indefintely. I patched that up, found the explicit exception, and tried to make the logic a little more clear.

Fixes # (issue)
#1262 , #1274 , #1283 

# Type of change

# Has your PR been tested?
caimanmanager test ran fine (demotest is currently running will report back if any problems). 

I pushed pretty hard against the `detrend_df_f()` function in the two main demo notebooks with different permutations of the kwargs. It behaved as expected with the logger outputting reasonable warnings (e.g., when using short time windows for percentiles).

### DIFF
--- a/caiman/utils/stats.py
+++ b/caiman/utils/stats.py
@@ -170,7 +170,7 @@ def df_percentile(inputData, axis=None):
     Used to determine the filtering level for DF/F extraction. Note that
     computation can be inaccurate for short traces.
 
-    If errors occur, will just return fallback value of median (50% percentile)
+    If errors occur, return fallback values
     """
     if axis is not None:
 
@@ -185,7 +185,7 @@ def df_percentile(inputData, axis=None):
         err = True
         err_count = 0
         max_err_count = 10
-        while err==True and err_count < max_err_count:
+        while err and err_count < max_err_count:
             try:
                 bandwidth, mesh, density, cdf = kde(inputData)
                 err = False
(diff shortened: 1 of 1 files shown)


caiman/utils/stats.py
```diff
@@ -169,6 +169,8 @@ def df_percentile(inputData, axis=None):
     Extracting the percentile of the data where the mode occurs and its value.
     Used to determine the filtering level for DF/F extraction. Note that
     computation can be inaccurate for short traces.
+
+    If errors occur, will just return fallback value of median (50% percentile)
```
pgunn, line 173: If errors occur, returns fallback value ...

caiman/utils/stats.py
```diff
@@ -181,24 +183,37 @@ def fnc(x):
     else:
         # Create the function that we can use for the half-sample mode
         err = True
-        while err:
+        err_count = 0
+        max_err_count = 10
+        while err==True and err_count < max_err_count:
```
pgunn, line 188: `while err and err_count`...

Testing boolean values for ==True is an antipattern

caiman/utils/stats.py
```diff
@@ -241,12 +254,13 @@ def kde(data, N=None, MIN=None, MAX=None):
     SqDCTData = (DCTData[1:] / 2)**2
 
     # The fixed point calculation finds the bandwidth = t_star
+    # fixed_point is function defined below
```
pgunn, line 257: what's this for?

pgunn: Oh, I just meant the comment

PR #1290, opened 2024-02-22
Move and rephrase info on caiman central

pgunn:
Testing README changes that make caiman central more prominent


EricThomson:
Seems like a good redesign (though maybe put below Quickstart Guide?).

Issue #1142, opened 2023-08-02
CaImAn 2023 Roadmap


EricThomson:
# Caiman Roadmap: August 2023 - August 2024
Our goals with Caiman naturally fall into three general categories:

- **Usability** – make Caiman's current feature base easier to use. 
- **Improvements** – improve or expand on Caiman's current feature base.
- **Community**: improve the quality of user experience, and increase base of users/contributors.

If you would like to get involved with any of the efforts in this roadmap, or have new ideas about how we can attain the above goals, please reach out by replying to this issue. This is an evolving document: check back to see how things are going, and if we have added any new goals. 

## Usability
Caiman already has many great features that we can do a better job bringing to existing and future users. Our roadmap for this:

- [ ] Update readme  [in progress]   
  - [X] Simplify install instructions (quickstart guide)    
  - [X] Add table of use cases with links to notebooks.    
  - [ ] Make caimanmanager usage the same across platforms.
  - [ ] Buttons: percent coverage, percent passing (see section on Improvements)    
- [ ] Rework CNMF demo notebook: `demo_notebook.ipynb` [in progress]
  - [ ] In general, simplify and make more beginner-friendly.
  - [ ] Discuss file types we can handle
  - [ ] Better address how to handle your own data.
  - [ ] Show how to iteratively search parameter space
  - [ ] Discuss loading estimates that have been saved
  - [ ] Add colab link
- [ ] Once `demo_notebook.ipynb` is done, use as template for others:
  - [ ] CNMFE demo
  - [ ] Motion correction demo
- [ ] Update documentation  
  - [ ] Docs are not rendering properly: fix (issue #997)
  - [ ] Update docs theme 
  - [ ] Update so docs are written in md instead of rst
  - [ ] Update docs: some are outdated or erroneous (gSiz)

## Improvements
While Caiman's current feature set is a strength, our goal is that it continuously evolve and improve. Our roadmap for this: 

- [ ] Add functionality to save estimates as csv (#1086).
- [ ] Get subidx parameter working with temporal cropping (issue #1106)
- [ ] Get rid of all pylab imports (good beginner issue)
- [ ] Add synthetic data generator class or function.
- [ ] Check number of frames for motion correction, with useful error messaging (#1036)
- [ ] Save to Pynapple format to work with Pynapple package (#1087).
- [ ] Refactor greedy initialization
  - [ ] Break up complex functions/methods into simple functions/methods
  - [ ] That are well-documented
  - [ ] And tested
[...]

PR #1302, opened 2024-03-21
Remove pylab imports

EricThomson:
# Description

Will close out item on Roadmap (#1142 ).  

Pylab is a relic of the days when Python was trying to be like Matlab and its use is strongly discouraged [even in its source code](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/pylab.py). This gets rid of it.

It includes things like:

```
from numpy import *
from numpy.fft import *
from numpy.random import *
from numpy.linalg import *
```

And it makes Caiman less readable with `pl.foo()` syntax whereas people expect `plt.foo()` when using the recommended api. 

# Type of change
- [X] Nnon-breaking change which fixes an issue

I'll be testing out things as I go through the different modules. 

One question: should we change the name of the back end in `movies.py` from `pylab` to `matplotlib`? It seems yes.  And, should we warn people? 

My guess is nobody uses it, I have tested it (it works):  the quality is horrible, we never recommend it -- in fact we literally throw a warning when people even *try* to use it. My inclination is just to make this change in the name of the back end. But if others thing we should give warnings at first, and change the name in a couple of cycles, I'd listen.

(Aside: I think a more important question is whether we even want this matplotlib interface for movies in there in the first place, but that's a bigger issue outside the scope of this PR. It seems like a maintenance burden not worth having.)

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -13,11 +13,11 @@ from IPython.display import display, Image
 import ipywidgets as widgets
 import logging
 import matplotlib
+import matplotlib.pyplot as plt
 import numpy as np
 import os
 import pathlib
 import pims
-import pylab as pl
 import scipy
 import skimage
 import sklearn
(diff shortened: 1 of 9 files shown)


pgunn (commented):
Generally looks good

EricThomson:
I believe that's everything: I changed the `backend` keyword from `pylab` to `pyplot` (using `matplotlib` seemed weird and this seems closer to the original, and will discourage its use hopefully: I do think it will be worth considering deprecating this option at some point in the future).

PR #1304, opened 2024-03-22
fix bug in correlation image calculation

EricThomson:
# Description

Fixes bug pointed out by @rcalfredson and implements fix suggested by @j-friedrich in discussion #1190 

# Type of change

# Has your PR been tested?
`caimanmanager test` passes. 

I ran:
`np.allclose(corr_im_local, corr_im_fft, rtol=1e-05, atol=1e-05, equal_nan=True)`

Where `corr_im_local` is the output of the buggy function `cm.local_correlations()`, and `corr_im_fft` is the output of the more canonical fft-based correlation function `cm.summary_images.local_correlations_fft()`

until it came back `True`. 

I haven't added this to the test suite. It might be a good idea though. 

### DIFF
--- a/caiman/summary_images.py
+++ b/caiman/summary_images.py
@@ -201,7 +201,7 @@ def local_correlations(Y, eight_neighbours: bool = True, swap_dim: bool = True, 
     Returns:
         rho: d1 x d2 [x d3] matrix, cross-correlation with adjacent pixels
     """
-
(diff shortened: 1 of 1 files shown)


EricThomson:
I will add test to this, but not sure when -- I'm on vacation most of this week, so let's do after upcoming release.

PR #1305, opened 2024-03-22
Fix problem with using reloaded cnmf object

### LINKED ISSUE #1264: Reloading cnmf object returns bytes rather than string
1. Operating System (Linux, **MacOS**, Windows):
 2. Hardware type (x86, **ARM**..) and RAM:
 3. Python Version (**3.11**):
 4. Caiman version (**1.9.16**):
 5. 
 6. How you installed Caiman (pure conda, conda + compile, colab, ..): conda and mamba
 7. Details: 

When reloading the CNMFE model using `load_CNMF`, the `cnmfe_model.params.data[fnames]` is of type numpy.bytes and requires a reset prior to changing parameters and evaluating components `cnmfe_model.params.data['fnames'] = [cnmfe_model.params.data['fnames'][0].decode('UTF-8')]`

EricThomson:
# Description
Issue raised with using reloaded data, the way things are serialized with hdf5, they are loaded as byte-encoded, so when you try to change parameters things don't work (e.g., filename validation throws an error). Rather than digging deeply into serialization in `recursively_load_dict_contents_from_group()` (or the corresponding save function), I added a check for the dtype in `check_consistency()` and if the filenames are not string, do the conversion. This is a bit hacky, but I tested it and it works 😬 

Maybe later we can do more internal validation in the load function, but for now I think this is a good workaround.

Thanks @JohnStout for pointing it out.

Closes #1264 

# Type of change

## Tests   
caimanmanager test works fine. I tested saving/loading and then changing parameters which was broken before, and now it works.

### DIFF
--- a/caiman/source_extraction/cnmf/params.py
+++ b/caiman/source_extraction/cnmf/params.py
@@ -927,8 +927,11 @@ class CNMFParams(object):
         if self.data['dims'] is None and self.data['fnames'] is not None:
             self.data['dims'] = caiman.base.movies.get_file_size(self.data['fnames'], var_name_hdf5=self.data['var_name_hdf5'])[0]
         if self.data['fnames'] is not None:
-            # following handles reloaded data where strings are byte-encoded
-            if not isinstance(self.data['fnames'][0], str):
+            # if fname was stored as string instead of list
+            if isinstance(self.data['fnames'], str):
+                self.data['fnames'] = [self.data['fnames']]
+            # convert reloaded data fnames from byte-encoded to string
+            if isinstance(self.data['fnames'][0], np.bytes_):
                 self.data['fnames'] = [fname.decode('utf-8') for fname in self.data['fnames']]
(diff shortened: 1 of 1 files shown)


pgunn:
This will only handle if it's bytes, but given that that's the most likely cause of this and there's no entirely general way to handle whatever else might be handled by check_consistency() it's a good solution for now.

We could in theory handle other things, like pathlib objects, but a lot would need to change if we ever wanted to support those.

EricThomson:
> This will only handle if it's bytes, but given that that's the most likely cause of this and there's no entirely general way to handle whatever else might be handled by check_consistency() it's a good solution for now.
> 
> We could in theory handle other things, like pathlib objects, but a lot would need to change if we ever wanted to support those.

It's definitely not perfect. This handles the current failure point (how filenames are handled by the hdf5 encoder/reader), but won't handle every possible problem. I'd tentatively place this in the "good enough for now" shed.

EricThomson:
I wasn't happy with checking for not string so I added explicit check for byte-encoding instead. Also, I had deleted the previous check for string as filename instead of list, so added that back in. 😬

Issue #1310, opened 2024-03-29
Demotest gives KeyError in cnmf/initialization.py


ethanbb:
Your setup:
 1. Operating System (Linux, MacOS, Windows): **Linux**
 2. Hardware type (x86, ARM..) and RAM: **x86, 128 GB**
 3. Python Version (e.g. 3.9): **3.11.8**
 4. Caiman version (e.g. 1.9.12): **`dev` branch**
 5. Which demo exhibits the problem (if applicable): **`demo_pipeline_cnmfE`**
 6. How you installed Caiman (pure conda, conda + compile, colab, ..): **conda + compile** (I think? conda and then editable install)
 7. Details: 

The demo `demo_pipeline_cnmfE` produces the following error. This seems to be due to #1239, and I believe the fix is to just update the name of the key in `greedyROI_corr` to `'temporal'`.

```
Traceback (most recent call last):
  File "/home/ethan/mambaforge/envs/caiman/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/ethan/mambaforge/envs/caiman/lib/python3.11/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
           ^^^^^^^^^^^^^^^^
  File "/synology/code/CalciumImaging/CaImAn/caiman/source_extraction/cnmf/map_reduce.py", line 113, in cnmf_patches
    cnm = cnm.fit(images)
          ^^^^^^^^^^^^^^^
  File "/synology/code/CalciumImaging/CaImAn/caiman/source_extraction/cnmf/cnmf.py", line 490, in fit
    self.initialize(Y)
  File "/synology/code/CalciumImaging/CaImAn/caiman/source_extraction/cnmf/cnmf.py", line 949, in initialize
    extra_1p = initialize_components(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/synology/code/CalciumImaging/CaImAn/caiman/source_extraction/cnmf/initialization.py", line 333, in initialize_components
    Ain, Cin, _, b_in, f_in, extra_1p = greedyROI_corr(
                                        ^^^^^^^^^^^^^^^
  File "/synology/code/CalciumImaging/CaImAn/caiman/source_extraction/cnmf/initialization.py", line 1171, in greedyROI_corr
    o = options['temporal_params'].copy()
[...]

PR #1311, opened 2024-03-29
Scanbox format updates

### LINKED ISSUE #1299: Suggestion: remove SBX loading code and recommend using sbxreader instead
### Discussed in https://github.com/flatironinstitute/CaImAn/discussions/1298

<div type='discussions-op-text'>

<sup>Originally posted by **ethanbb** March 14, 2024</sup>
Hello,

This is mainly a documentation suggestion. My microscope saves Scanbox files, and I saw CaImAn has support for loading them, so I tried that but got an error because the loading functions are not up-to-date with newer versions of the file format (which has changed several times). The specific error was about inferring the number of channels (my `info` did not have `'channels'`, but rather the newer `'chan'` field).

Looking into it a little more, I realized that the current function doesn't account for a number of features, including:
- Volumetric scanning (optotune)
- Bidirectional scanning
- "Fold lines" (multiple subframes per frame)
- Loading channels other than the first one

I took a few days to implement these features, and I was going to submit a pull request. However, towards the end I realized that the package [sbxreader](https://github.com/jcouto/sbxreader) exists (which suite2p uses for loading). Besides being actively maintained, it's much better than what I wrote because it memory-maps the file - I realized when trying to test my function that it was not going to happen because my 40-minute recording was too large to fit into memory. It also has another function for loading all metadata, which is probably important to almost everyone who would need to load SBX data.

I could continue to tweak what I wrote to use memmap etc., but I think what makes the most sense is to just recommend Scanbox users to install this package and use it to get an np.memmap of their data, then load it into CaImAn using the movie constructor directly. To encourage this (and because it's out of date), it probably makes sense to remove the current SBX loading code.</div>

ethanbb:
# Description

- Updates .sbx loading code to deal with newer versions of the format;
- Adds functions to convert .sbx files and chains thereof to .tifs for more flexibility;
- Adds a function to load the metadata;
- Pulls all the sbx-relevant code into a separate file under `utils`.

Fixes #1299 

# Type of change

# Has your PR been tested?

Yes, tests for .sbx format added to `tests/test_sbx.py`, using example datafiles added to `testdata`
All tests passed with `caimanmanager test`. `caimanmanager demotest` fails because of #1310.


caiman/base/movies.py
```diff
@@ -35,6 +35,7 @@
 import caiman.mmapping
 import caiman.summary_images
 import caiman.utils.visualization
+import caiman.utils.sbx_utils
```
pgunn, line 38: Can we flip the order to keep these alphabetised?

caiman/base/movies.py
```diff
@@ -1575,10 +1576,11 @@ def load(file_name: Union[str, list[str]],
 
         elif extension == '.sbx':
             logging.debug('sbx')
-            if subindices is not None:
-                return movie(sbxreadskip(file_name[:-4], subindices), fr=fr).astype(outtype)
-            else:
-                return movie(sbxread(file_name[:-4], k=0, n_frames=np.inf), fr=fr).astype(outtype)
+            meta_data = caiman.utils.sbx_utils.sbx_meta_data(file_name[:-4])
```
pgunn, line 1579: We have a lot of code throughout the codebase that does this; eventually we'll want to convert them all to use os.path.splitext() instead of indexing into the string. When I fix them all I'll grep for stuff like this, so it's fine to leave this in (although if you want to adjust it yourself you can)

caiman/tests/test_sbx.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python
+
+import os
+
+import numpy as np
+import numpy.testing as npt
```
pgunn, line 6: Can we alphabetise these parts of the imports? (the caiman imports going later is fine

caiman/utils/sbx_utils.py
```diff
@@ -0,0 +1,525 @@
...
+Utility functions for Neurolabware Scanbox files (.sbx)
+"""
+
+import os
+import logging
+from typing import Iterable
+
+import numpy as np
+import scipy
+import tifffile
```
pgunn, line 13: Can we alphabetise these imports?

pgunn (commented):
Very nice. I particularly like your type annotations style; might try out some of that with my own code because of how clean it looks. Left a few little comments.

pgunn:
Looking good, will merge once CI finishes. Thanks!

pgunn:
@ethanbb Ah, I'm fixing CI now and it turns out CI found an issue with the code:

```
  File "/home/runner/work/CaImAn/CaImAn/caiman/base/movies.py", line 37, in <module>
    import caiman.utils.sbx_utils
  File "/home/runner/work/CaImAn/CaImAn/caiman/utils/sbx_utils.py", line 435
    return chunk[:, *np.ix_(*subindices[1:])]
                    ^
SyntaxError: invalid syntax
Error: Process completed with exit code 1.
```

pgunn:
Looks like this is 3.11-specific syntax; is there a chance you can, for now, rephrase it to be compatible back to 3.9?

ethanbb:
Yup, no problem.

PR #1313, opened 2024-03-29
Test ci fixes

### LINKED ISSUE #1312: CI is broken, Windows-WSL (and some other Linux) users need some extra systems packages
Very recently, we've seen dependencies on OpenGL system packages turning up in both Github-CI and for Ubuntu-on-WSL users. The former is broken, the latter is fixable with some extra steps.

The dependency is coming in through OpenCV, but it's not their fault; unfortunately they recently changed dependencies to Qt6, which for some reason needs libEGL. This is tracked upstream here:

https://github.com/conda-forge/opencv-feedstock/issues/401

The suggested solution - depending on headless build variants of opencv, won't work for us (because I think we use opengl to play things occasionally).

We either need:
A) To convince opencv to not build against Qt6
B) To convince the Qt6 developers to remove dependencies on OpenGL
C) To somehow handle getting WSL users to install some additional system depencies (annoying but not hard) and to get Github's workers to install opengl libraries (no idea what's involved, may or may not be possible) or to have a different yaml file for CI (really don't want to do this)

Investigating all of these.

pgunn:
Hoping this will fix #1312
Let's see how it goes

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -29,6 +29,11 @@ jobs:
           environment-file: environment.yml
           activate-environment: caiman
 
+      - name: Install OS Dependencies
+        shell: bash -l {0}
+        run: |
+          sudo apt-get install libglapi-mesa libegl-mesa0 libegl1 libopengl0 libgl1-mesa-glx
+
       - name: Install Dependencies
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


pgunn:
Looks like we're allowed to use sudo, and this got CI working again.

PR #1314, opened 2024-03-29
Fix incompatibilities in test_sbx for python <3.11

ethanbb:
# Description

Fixes build error in #1311 

# Type of change

# Has your PR been tested?

```caimanmanager test``` passes on 3.11 - if any invalid syntax for <3.11 remains, should be caught in CI

### DIFF
--- a/caiman/utils/sbx_utils.py
+++ b/caiman/utils/sbx_utils.py
@@ -9,11 +9,11 @@ import numpy as np
 import os
 import scipy
 import tifffile
-from typing import Iterable
+from typing import Iterable, Union, Optional
 
-DimSubindices = Iterable[int] | slice
-FileSubindices = DimSubindices | Iterable[DimSubindices]    # can have inds for just frames or also for y, x, z
-ChainSubindices = FileSubindices | Iterable[FileSubindices] # one to apply to each file, or separate for each file
+DimSubindices = Union[Iterable[int], slice]
+FileSubindices = Union[DimSubindices, Iterable[DimSubindices]]  # can have inds for just frames or also for y, x, z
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ethanbb** [comment]
Well looks like the nice type annotation syntax doesn't work on 3.9... updating now

**Comment by pgunn** [comment]
Looks like the type annotation syntax is also not compatible with 3.9:
```
  File "/home/runner/work/CaImAn/CaImAn/caiman/utils/sbx_utils.py", line 14, in <module>
    DimSubindices = Iterable[int] | slice
TypeError: unsupported operand type(s) for |: '_GenericAlias' and 'type'
```

We're looking forward to pushing our python dependency to 3.10-3.12 for a lot of reasons, but we're not quite ready to do so yet.

pgunn:
At least stuff like this reminds us that python is getting better! I think there were just some upstream packaging issues that currently block us from switching over (tensorflow and one other thing, I think); when we do I'm looking forward to going through the codebase and modernising the syntax (my favourite kind of diff to work on)

pgunn:
Thanks! I'll be cutting a new release shortly (today or tomorrow); mostly a bugfix for the thing you found earlier that I thought was already out, but your sbx fix here will also be part of it; you won't need to use a compiled build for very long (unless you want to).

ethanbb:
Great, thank you for being so responsive and helping me get this out!

kushalkolar:
I'm think this is py3.10+ https://peps.python.org/pep-0604/ 

according to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) python3.9 support stops in numpy on April 5 so it's not too far out, but tensorflow usually lags behind quite a bit.

PR #1316, opened 2024-04-01
SBX utils: ensure correct photometric setting for TIFs and allow extracting single plane

ethanbb:
# Description

This PR includes two changes:
- A bugfix for the condition when a 3D SBX with 3 or 4 planes is converted to TIF. By default, tifffile assumes the 4th dimension is the color channel, so it is necessary to set `photometric` to `MINISBLACK` to avoid each plane being 3-dimensional.
- An added feature, the `plane` argument to SBX loading/conversion functions, which allows extraction of a single plane without keeping the singleton dimension. This is useful when opting to process each plane of a volumetric dataset individually.

# Type of change

# Has your PR been tested?

Yes, ```caimanmanager test``` passes and new tests have been added for both of these changes.

### DIFF
--- a/caiman/tests/test_sbx.py
+++ b/caiman/tests/test_sbx.py
@@ -3,12 +3,15 @@
 import numpy as np
 import numpy.testing as npt
 import os
+import tifffile
 
 import caiman as cm
 from caiman.paths import caiman_datadir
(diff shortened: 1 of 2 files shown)


pgunn:
Looks good. Thanks!

Issue #1332, opened 2024-04-13
CNN model cannot be loaded with latest `keras` and `tensorflow`


kushalkolar:
1. Operating System (Linux, MacOS, Windows): Debian 11
 2. Hardware type (x86, ARM..) and RAM: x86_64, Ryzen 5900X
 3. Python Version (e.g. 3.9): 3.11
 4. Caiman version (e.g. 1.9.12): 1.10.1

Component evaluation fails with latest tensorflow & keras since it cannot load the CNN model file, `cnn_model.json`. This is with keras [v3.2.1](https://github.com/keras-team/keras/releases/tag/v3.2.1) which was released 3 days ago and tensorflow [v2.16.1](https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1) released on March 7.

This does not occur with keras v2.12.0 and tensorflow v2.12.0 (I'm not sure what's the last version in which the models can be loaded). Rolling back to these versions also allows the model file to be loaded.

Here is the traceback in-place when you try to use evaluate_components in caiman:

```python
File "[/home/kushal/repos/CaImAn/caiman/source_extraction/cnmf/estimates.py", line 1036](http://localhost:8888/lab/tree/repos/mesmerize-core/notebooks/repos/CaImAn/caiman/source_extraction/cnmf/estimates.py#line=1035), in evaluate_components
    estimate_components_quality_auto(imgs, self.A, self.C, self.b, self.f, self.YrA,
  File "[/home/kushal/repos/CaImAn/caiman/components_evaluation.py", line 608](http://localhost:8888/lab/tree/repos/mesmerize-core/notebooks/repos/CaImAn/caiman/components_evaluation.py#line=607), in estimate_components_quality_auto
    idx_components, idx_components_bad, cnn_values = select_components_from_metrics(A, dims, gSig, r_values, comp_SNR,
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[/home/kushal/repos/CaImAn/caiman/components_evaluation.py", line 653](http://localhost:8888/lab/tree/repos/mesmerize-core/notebooks/repos/CaImAn/caiman/components_evaluation.py#line=652), in select_components_from_metrics
    predictions, _ = evaluate_components_CNN(A, dims, gSig)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1334, opened 2024-04-23
environment.yml: handle variant and newer condas better

pgunn:
Be smarter about alternate builds of conda, set channel-priority to handle if user/distro has set it to something else.

This is an attempt to improve on the install issues seen with #1332 

Note that the `channel_priority: true` may seem wrong, but it's a ternary value, with false meaning disabled, true meaning flexible, and strict (a new value) meaning strict. We don't want strict, at least as long as we need tensorflow on Windows. Flexible should do the job.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
-          environment-file: environment.yml
+          environment-file: environment-minimal.yml
           activate-environment: caiman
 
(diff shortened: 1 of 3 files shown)


pgunn:
Actually channel_priority cannot be set in environment.yml; I misread some docs. Removed that bit.

ethanbb:
Yes, I confirmed this works for me with Miniforge on Windows, so no need to install tensorflow separately after installing everything else from conda-forge. Also, it seems that Miniforge at least defaults to flexible channel priority, so that shouldn't be a problem (I _think_ that's by far the most popular distribution that doesn't include `defaults`).

Issue #1336, opened 2024-04-26
Matlab support broken due to changes in scipy


pgunn:
scipy removed _open_file from scipy.io.matlab.mio, and that entire namespace is going away in SciPy 2.0.0. We were using this to handle reading .mat files (both old-style ones and new-style hdf5 ones)

The most obvious replacement, scipy.io.matlab.loadmat(), throws an exception rather than letting us read versions from attributes, as so:

```
>>> fh = scipy.io.matlab.loadmat('example_movies/blood_vessel_10Hz.mat')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/home/pgunn/miniconda3/envs/caibuild/lib/python3.11/site-packages/scipy/io/matlab/_mio.py", line 226, in loadmat
    MR, _ = mat_reader_factory(f, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/home/pgunn/miniconda3/envs/caibuild/lib/python3.11/site-packages/scipy/io/matlab/_mio.py", line 80, in mat_reader_factory
    raise NotImplementedError('Please use HDF reader for matlab v7.3 '
NotImplementedError: Please use HDF reader for matlab v7.3 files, e.g. h5py
```

This probably won't be that hard to handle (and in some ways probably mirrors some of the recent sbx improvements @ethanbb did for us), unless we bump into deeper API changes.

Currently demo_Ring_CNN is broken because of these changes.

pgunn:
It looks like the current codebase was able to load() a traditional .mat file, but it was not able to get_file_size() on it, meaning it would take unusually fully parameterised cnmfparams for this to have worked (even before the current issue). Instead it usually just acted as a way to quickly push the user off to using hdf5 support if it were in a modern not-really-a-mat format.

There's a working old-style datafile on our clustered filesystem here:
/mnt/ceph/data/neuro/VariousLabeling/Jan_ROIS/Jan40_exp2_001_ROIs.mat

We could decide to stop supporting classic mat files (only supporting the new hdf5 style), and simplify the code a lot, or we could rebuild classic mat support using the newer APIs in scipy; whatever we do we need to not use the old scipy APIs as they no longer work.

pgunn:
[...]

PR #1337, opened 2024-04-26
fix mat support

pgunn:
This fixes use of scipy APIs that are no longer in scipy to read old-style .mat files. ( #1336 )

Support for the newer hdf5-based mat format is retained (and by newer we're talking about a change made a long time ago).

There were two other ways to fix the import problem:
1) Have a slightly friendlier user message for people trying to load old mat files, by doing try/except in a few places, catching the exception, and guessing that they were trying to load the earlier format and tell them.
2) Attempt to use scipy.io.matlab.loadmat() instead, building support for that back into the codebase (classic mat handling had been broken for awhile and would've needed some work)

I decided not to do the first because it complicates some already complicated code for the sake of an obsolete format (and the try/exception model actually wouldn't tell us for sure why an OSError was thrown so it could be confusing no matter what).
I decided not to do the second because it's a bit of work for a format people shouldn't be using anymore and it's something we may have decided to stop supporting anyhow even if the code was working (unless users told us they need it).

If some users of old-style mat files turn up and tell us they really need the old format (even though it hasn't worked in caiman for several releases at least), I can move to the second solution, but I'm otherwise happy with this choice.

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -1244,7 +1244,8 @@ def load(file_name: Union[str, list[str]],
             dimension of the movie along x and y if loading from a two dimensional numpy array
 
         var_name_hdf5: str
-            if loading from hdf5/n5 name of the dataset inside the file to load (ignored if the file only has one dataset)
+            if loading from hdf5/n5 name of the dataset inside the file to load (ignored if the file only has one dataset).
+            This is also used for (new-style) mat files
 
         in_memory: bool=False
             This changes the behaviour of the function for npy files to be a readwrite rather than readonly memmap,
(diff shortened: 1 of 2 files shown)


pgunn:
Ugh, I based this off the wrong branch. Redoing the diff.

ethanbb:
I would mildly favor at least catching whatever error gets thrown for older .mat files and erroring with a message that suggests that as a possible cause of the error, given how like I said the older format is still the default. But I do support dropping any appearance of real support for it.

pgunn:
Hm. Okay. The code for that isn't too bad/bulky.

ethanbb:
Thanks! Yeah, the days of intentionally saving non-HDF5 mat files are long gone (for everyone I know), but the days of accidentally ending up with them are unfortunately still with us.

PR #1353, opened 2024-05-16
Fix medw loading as b'NoneType' instead of None

ethanbb:
# Description

I encountered a bug when running a CNMF refit after loading the CNMF object from HDF5. This was caused by `params.spatial['medw']` being `b'NoneType'` rather than `None`.

`None` parameters are currently converted to the string 'NoneType' during saving, which is read back into Python as `b'NoneType'`, and this should be converted back to `None`, but due to the way the conditionals are structured, parameters which should be converted to tuples if not None are not converted if they equal `b'NoneType'`. (They are converted if they are the regular string `'NoneType'`.) I fixed it so that both NoneType cases are handled first, and otherwise simplified the conditional structure (but that should be the only change in behavior).

Partially addresses #1264.

# Type of change

# Has your PR been tested?

Yes, `caimanmanager test` and `caimanmanager demotest` pass.

### DIFF
--- a/caiman/utils/utils.py
+++ b/caiman/utils/utils.py
@@ -562,7 +562,7 @@ def recursively_load_dict_contents_from_group(h5file:h5py.File, path:str) -> dic
     for key, item in h5file[path].items():
         if isinstance(item, h5py._hl.dataset.Dataset):
             val = item[()]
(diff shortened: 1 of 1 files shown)


pgunn:
Looks good, thanks!

PR #1363, opened 2024-06-13
Fix removeBL for 3D movies

ethanbb:
# Description

This PR is mainly focused on making sure `local_correlations_movie` works for 3D movies, by adding an `is3D` flag for passing to `caiman.load` and also slightly modifying `movie.removeBL`. I also fixed other occurrences in local_correlations.py where `caiman.load` is used without an `is3D` flag; however, I have not tested whether all these functions actually work for 3D movies.

This fix is necessary to get mesmerize-core's motion correction routine to work for 3D data. But I'd understand if you want to make it a little more broad.

# Type of change

# Has your PR been tested?

```caimanmanager test``` and ```caimanmanager demotest``` pass. I tested manually by running some of my data through `local_correlations_movie` (both from mesmerize-core and otherwise).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,8 +52,7 @@ __pycache__/
 
 # C extensions
 *.so
-*.c
-*.cpp
+caiman/source_extraction/cnmf/oasis.cpp
 
 #movie tif
 *.tif
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,7 @@ __pycache__/
 # C extensions
 *.so
 *.c
+*.cpp
 
 #movie tif
 *.tif
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -552,15 +552,15 @@ class movie(caiman.base.timeseries.timeseries):
     
     def to2DPixelxTime(self, order='F'):
         """
-        Transform 3D movie into 2D
+        Transform 3D or 4D movie into 2D
         """
-        return self.transpose([2,1,0]).reshape((-1,self.shape[0]),order=order)  
+        return self.transpose().reshape((-1,self.shape[0]),order=order)
 
     def to3DFromPixelxTime(self, shape, order='F'):
         """
-            Transform 2D movie into 3D
+            Transform 2D movie into 3D or 4D
         """
-        return to_3D(self,shape[::-1],order=order).transpose([2,1,0])
+        return to_3D(self,shape[::-1],order=order).transpose()
     
     def computeDFF(self, secsWindow: int = 5, quantilMin: int = 8, method: str = 'only_baseline', in_place: bool = False,
                    order: str = 'F') -> tuple[Any, Any]:
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -550,17 +550,17 @@ class movie(caiman.base.timeseries.timeseries):
             self -= caiman.movie(cv2.resize(res,pixs.shape[::-1]),fr=self.fr).to3DFromPixelxTime(self.shape) 
             return self
     
-    def to2DPixelxTime(self, order='F'):
+    def to2DPixelxTime(self, order='F') -> 'movie':
         """
-        Transform 3D or 4D movie into 2D
+        Transform 3D or 4D movie into 2D (pixels x time)
         """
-        return self.transpose().reshape((-1,self.shape[0]),order=order)
+        return self.T.reshape((-1,self.shape[0]),order=order)
 
-    def to3DFromPixelxTime(self, shape, order='F'):
+    def to3DFromPixelxTime(self, shape, order='F') -> 'movie':
         """
-            Transform 2D movie into 3D or 4D
+        Transform 2D movie (pixels x time) into 3D or 4D
         """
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by pgunn** [comment]
It looks generally workable; I wonder if this might be improved by modifying the params object to notice and activate this automatically when the passed in dims has more cardinality (data.dims); right now there's some design tension between with passing params around and having it control behaviour versus passing parameters to various functions (this is a longer-term concern though).

I worry about the code being fragile to future refactoring without our noticing it.
If you can add a test that would exercise this functionality, that'd be good.

**Comment by ethanbb** [comment]
You know what, I don't think the `is3D` flag is necessary after all. I just looked again at the load function and realized that it's only used if the file name passed in is actually a list. This is happening in mesmerize-core, but it's against the documented api of `local_correlations_movie_offline`, so that call should just be changed.

The fixes to those reshaping utilities are still necessary, though. I can add tests for them.

File: .gitignore [hunk 1]
```diff
@@ -53,6 +53,7 @@ __pycache__/
 # C extensions
 *.so
 *.c
+*.cpp
```
ethanbb, line 56: snuck this in b/c git kept wanting to add oasis.cpp

ethanbb:
Also, slightly off-topic now but:

```python
if channel is not None:
    logging.debug(m.shape)
    m = m[channel].squeeze()
    logging.debug(f"Movie shape: {m.shape}")
```

is this clause in `load_movie_chain` ever used at this point? Since as I understand the output of `load` should not have a channel dimension (first dimension of any `movie` should be time, right?) And if not, if we're never squeezing the movie, do we need `is3D` at all; can't we infer it from the dimensionality of the movie(s)?

.gitignore
pgunn: We'll need to remember this if we ever add real cpp files to the project, but this is good for now.

kushalkolar: put the full path to `oasis.cpp`?

ethanbb: If we're doing that, since this is the only C extension, we could also remove `*.c` for consistency?

pgunn: The full path is a good idea. How much we use wildcards isn't something I care much about; I suspect the initial form of that file probably came from a template someone grabbed way back.

ethanbb: Sorry I'm a little confused - what should I update this to exactly?

pgunn: It's fine to remove the *.c if you want, and to refactor the cpp filter to be only pathed rather than general. Up to you.

pgunn:
> Also, slightly off-topic now but:
> 
> ```python
> if channel is not None:
>     logging.debug(m.shape)
>     m = m[channel].squeeze()
>     logging.debug(f"Movie shape: {m.shape}")
> ```
> 
> is this clause in `load_movie_chain` ever used at this point? Since as I understand the output of `load` should not have a channel dimension (first dimension of any `movie` should be time, right?) 

We seem to have no current code that uses channel; I'm guessing it was added to support some people who had their data in an unusual format, and we probably should tell people in that case to convert first. I'm going to look back through repo history to see when this was added (assuming it wasn't there from the start) to get a bit more history.

pgunn:
> > Also, slightly off-topic now but:
> > ```python
> > if channel is not None:
> >     logging.debug(m.shape)
> >     m = m[channel].squeeze()
> >     logging.debug(f"Movie shape: {m.shape}")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is this clause in `load_movie_chain` ever used at this point? Since as I understand the output of `load` should not have a channel dimension (first dimension of any `movie` should be time, right?)
> 
> We seem to have no current code that uses channel; I'm guessing it was added to support some people who had their data in an unusual format, and we probably should tell people in that case to convert first. I'm going to look back through repo history to see when this was added (assuming it wasn't there from the start) to get a bit more history.

The channel argument was added to load_movie_chain() on 18 September 2017, in a large diff called "pre merging 1photon with master"; ( 02d7977 ) this was probably a squashed merge so we don't have history for it.

The channel argument was added to load() on 8 December 2017 in a diff called "multiple file support" that made direct use of load_movie_chain() usually unnecessary because load() would call it for you if you passed a list (pity we never took that principle all the way; maybe something for the future).

I can't find any code that uses the channel argument to either function in the repo, and it was never documented, so we probably can remove it.

ethanbb:
Hey, thanks for taking a look. I more recently made a discussion #1366 about this channel argument - instead of removing, maybe we should make it more functional? Or what is the alternative for multi-channel recordings?

pgunn:
> Hey, thanks for taking a look. I more recently made a discussion #1366 about this channel argument - instead of removing, maybe we should make it more functional? Or what is the alternative for multi-channel recordings?

I think the alternative would probably be for people to pre-convert their data out of such formats.

If you'd like to convert it to behave differently, I'm open to that; otherwise I'm inclined to just remove it.

ethanbb:
@pgunn I think the discussion got a little off track... this is a pretty minor change, do you think we could push this through soon to unblock the corresponding issue in mesmerize-core?
Thanks!

pgunn:
Sure

Issue #1376, opened 2024-07-18
Fine-grained logging, and/or CNMF.fit() logging adjustments


pgunn:
This is a tracking issue to either/both:

1) Revisit the level of noise that the logging functions in CNMF.fit() can produce (or revisit the levels), and
2) To consider implementing named loggers in more of the code to give users more control over what logs they want to see

If addressing the first produces satisfactory results, we may not need to do the second (which has a complexity burden for users and we'd need to document it - begins to seem like overengineering).

(inspired by #1374)

pgunn:
Another option:
Put all caiman logging code under a single named logger that's not the global one, to let users have whatever logging they want in their own code that's distinct from how caiman logs.

pgunn:
Going with that latter option.

pgunn:
Merged and in the most recent release (will be on conda-forgen shortly)

Issue #1374, opened 2024-07-18
How to disable logging?


nick-youngblut:
I want to use logging in my script that runs caiman, but I do not want the excessive logging output from `CNMF.fit()`.  According to:

```
print(f"Name: {__name__}")
logger = logging.getLogger(__name__)
```

in `cluster.py`, the `Name: caiman.cluster`, so I set the following at the start of my script:

```
logging.getLogger("caiman.cluster").setLevel(logging.ERROR)
```

...but I still get the full logging output from `CNMF.fit()`

I've also tried:

```
logging.disable(logging.CRITICAL + 1)
```

...but I still get the full logging output.

How does one disable the verbose output from 

### Setup

 1. Operating System: Ubuntu 22.04.4
 2. Hardware type (x86, ARM..) and RAM: x86, 1 Tb
 3. Python Version (e.g. 3.9): 3.11.9
 4. Caiman version (e.g. 1.9.12):  1.11.1
 5. Which demo exhibits the problem (if applicable): N/A
 6. How you installed Caiman (pure conda, conda + compile, colab, ..): mamba

pgunn:
I'll look into this; the codebase used to have snippets all over the place that reset the loglevel; I've removed a lot of them but I may have missed some.

pgunn:
I just read your question more carefully; I think you're asking for the ability to set different logging levels for different parts of the code; our support for this is pretty halfhearted right now (caiman.cluster will only affect things in that part of the code, I think.

Right now only caiman.utils.labelling and caiman.cluster provide that kind of fine-grained control.

For the rest of the code, you'll just need to set up the general logger to a certain level and get exactly that level.

I'll file an issue to take a look at this in the future (and in general to review the level of noise that CNMF.fit() produces)

nick-youngblut:
To be clear, I just want to turn off logging (at least disable INFO messages) for caiman, but I want logging for the rest of my script that does more than just runs `CNMF.fit()`

pgunn:
Right now the best way to do that is a bit unfortunate; you'll want to save the state of the global logger, change the global log level just before calling caiman code, and then change it back afterwards.

This is a good use case for me to think about in the future though; if we put all caiman code under one logger that's not the global logger, we can document that almost as easily as changing the global logger settings, and it would give you the flexibility you want.

Thoughts?

pgunn:
@nick-youngblut If you can give me your thoughts on the idea above..

nick-youngblut:
[...]

PR #1378, opened 2024-07-24
Logging cleanup

pgunn:
This implements #1376 and addresses #1374

It moves almost all use of logging to use a named logger rather than the base logger. This allows code that calls caiman to have different log levels than caiman does, giving the user more control over logging.

I included some minor cleanups of things I saw along the way (like f-string conversions and import cleanups), and updated the docs on the logger (which were incorrect even before this diff; they suggested a logger format that's not valid)

As I write this:
1) I need to modify the jupyter notebooks too
2) Need to see if this passes CI
3) Will need to look at the diff a little more to make sure I didn't miss adding the code to reach the named logger I added to functions/methods

I'll push another revision to this to deal with the first, at least, and anything I need to change for the second and third


pgunn:
Second and third concerns are taken care of. I still need to loop back and adjust the jupyter notebooks. Will do that either later tonight or tomorrow.

PR #1385, opened 2024-08-09
Fix error when saving array with NaNs (occurs with corr_img when use_corr_img is True)

ethanbb:
# Description

The function `recursively_save_dict_contents_to_group` checks that an ndarray was written correctly to the HDF5 file by calling `np.array_equal`, but it currently does not set the option `equal_nan=True`, meaning that if there are any NaNs in the array, it returns false. Regardless of whether NaNs are expected, this is probably not what we want if we are just doing a fidelity check.

I initially tried just changing this to true, but apparently it can fail with a TypeError ("ufunc 'isnan' not supported for the input types") in some cases. So I changed it to only be true if the ndarray is of a floating-point type, which seems to work.

# Type of change

# Has your PR been tested?

Have not tested beyond this; may be good to add a unit test for it, which I can come back and do later.

### DIFF
--- a/caiman/utils/utils.py
+++ b/caiman/utils/utils.py
@@ -516,7 +516,7 @@ def recursively_save_dict_contents_to_group(h5file:h5py.File, path:str, dic:dict
             except:
                 item = np.array(item).astype('|S32')
                 h5file[path + key] = item
-            if not np.array_equal(h5file[path + key][()], item):
+            if not np.array_equal(h5file[path + key][()], item, equal_nan=item.dtype == 'f'):  # just using True gives "ufunc 'isnan' not supported for the input types"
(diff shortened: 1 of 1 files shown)


pgunn:
Looks great, thanks!

ethanbb:
Sorry, this has a typo. It should be `item.dtype.kind == 'f'`. But now I'm confused why it fixed the error I was getting, so looking into that...

pgunn:
Ah, I may have merged too soon. If you want to followup with another PR..

ethanbb:
OK got it, so `dtype('float32')` compares equal with 'f', which is why it worked for me, but other floating-point dtypes do not. So the right thing to do is to use the `kind`.

Issue #1185, opened 2023-10-11
Weird errors when CNMFE params are quite off


kushalkolar:
Troubleshooting mescore CI and ran into this, reproduced on CNMFE demo nb by setting `gSig = (10, 10)` and `gSiz = (41, 41)`, everything else is the same. These params are quite far from ideal for this movie, but it's weird that a pickling error is thrown: 

The reason is because `gSiz > rf`, but this type of error is very cryptic. PR incoming...

Side note, `dview.map_async(cnmf_patches, args_in).get(4294967)` is retrieving chunks of size 2^32 - 1, curious why that's the chunk size? 

```python
---------------------------------------------------------------------------
MaybeEncodingError                        Traceback (most recent call last)
Cell In[32], line 2
      1 cnm = cnmf.CNMF(n_processes=n_processes, dview=dview, Ain=Ain, params=opts)
----> 2 cnm.fit(images)

File ~/repos/CaImAn/caiman/source_extraction/cnmf/cnmf.py:594, in CNMF.fit(self, images, indices)
    589 if not isinstance(images, np.memmap):
    590     raise Exception(
    591         'You need to provide a memory mapped file as input if you use patches!!')
    593 self.estimates.A, self.estimates.C, self.estimates.YrA, self.estimates.b, self.estimates.f, \
--> 594     self.estimates.sn, self.estimates.optional_outputs = run_CNMF_patches(
    595         images.filename, self.dims + (T,), self.params,
    596         dview=self.dview, memory_fact=self.params.get('patch', 'memory_fact'),
    597         gnb=self.params.get('init', 'nb'), border_pix=self.params.get('patch', 'border_pix'),
    598         low_rank_background=self.params.get('patch', 'low_rank_background'),
    599         del_duplicates=self.params.get('patch', 'del_duplicates'),
    600         indices=indices)
    602 self.estimates.bl, self.estimates.c1, self.estimates.g, self.estimates.neurons_sn = None, None, None, None
    603 logging.info("merging")

File ~/repos/CaImAn/caiman/source_extraction/cnmf/map_reduce.py:250, in run_CNMF_patches(file_name, shape, params, gnb, dview, memory_fact, border_pix, low_rank_background, del_duplicates, indices)
    248 if dview is not None:
    249     if 'multiprocessing' in str(type(dview)):
--> 250         file_res = dview.map_async(cnmf_patches, args_in).get(4294967)
    251     else:
    252         try:

[...]

PR #1387, opened 2024-08-11
Fix conversion from sparse matrix to dense array when nb == 0 and not low_rank_background

ethanbb:
# Description

Fixes a spot in the code where a sparse matrix was being passed into `np.array`, which is not allowed and throws an error. This block of code seems to execute when low_rank_background is False and nb is not set to 0. 

Partially addresses #1185, nel-lab/mesmerize-core#231.

# Type of change

# Has your PR been tested?

I made some changes to test_toydata.py so that it can more easily test multiple different sets of parameters, and then I added a couple of sets for batch processing. I confirmed that without this fix, the one called 'patch-not-lowrank' reproduces this error, and with it the tests pass.


ethanbb:
Bumping @pgunn, related to a CI error in mesmerize (and this is a bug that people might run into). 

I realize this mostly became a change to test_toydata, which I hope was an improvement, but I'm definitely open to doing that differently.

PR #1390, opened 2024-08-16
caiman core

kushalkolar:
Just trying to keep an updated branch of caiman with only the core algos, no bokeh, no nwb, etc. I don't intend to merge it, maybe it can become a separate repo at some point.


kushalkolar:
ok seems like I can't use github's merge conflict resolver here so there isn't much point in me keeping this as a PR here.

PR #1410, opened 2024-10-01
Fix docstrings important for seeded initialization

kushalkolar:
I spent a few hours figuring this out so it should be properly documented without having to run the notebook and figure out the exact types that are accepted.


kushalkolar:
good to go

PR #1422, opened 2024-11-14
Include background component indices when removing empty/nan components in spatial.py

ethanbb:
# Description

This fixes an indexing error in the "Eliminating empty and nan components" step that occurs when using seeded CNMF. The error I was getting is as follows:

```python
  File "/u/ethan/mesmerize-core/mesmerize_core/algorithms/cnmf.py", line 94, in run_algo
    cnm = cnm.fit(images)
          ^^^^^^^^^^^^^^^
  File "/u/ethan/CaImAn/caiman/source_extraction/cnmf/cnmf.py", line 523, in fit
    self.update_spatial(Yr, use_init=True)
  File "/u/ethan/CaImAn/caiman/source_extraction/cnmf/cnmf.py", line 915, in update_spatial
    update_spatial_components(Y, C=self.estimates.C, f=self.estimates.f, A_in=self.estimates.A,
  File "/u/ethan/CaImAn/caiman/source_extraction/cnmf/spatial.py", line 189, in update_spatial_components
    ind2_ = [ind_list[np.setdiff1d(a,ff)] if len(a) else a for a in ind2_]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/ethan/CaImAn/caiman/source_extraction/cnmf/spatial.py", line 189, in <listcomp>
    ind2_ = [ind_list[np.setdiff1d(a,ff)] if len(a) else a for a in ind2_]
             ~~~~~~~~^^^^^^^^^^^^^^^^^^^^
IndexError: index 1637 is out of bounds for axis 0 with size 1637
```

It looks like this is happening because when `A_in` is provided and has type bool, the following lines add indices corresponding to the spatial components (>= `nr`) to non-empty entries of `ind2_`:

https://github.com/flatironinstitute/CaImAn/blob/ce2cb179cb801c3541f40d0c370ae961381acf34/caiman/source_extraction/cnmf/spatial.py#L1066-L1067

However, the code to remove empty components assumes that the highest original component number is `nr-1`: 

https://github.com/flatironinstitute/CaImAn/blob/ce2cb179cb801c3541f40d0c370ae961381acf34/caiman/source_extraction/cnmf/spatial.py#L183-L187

To fix this, I changed this to use `nr + nb`. However, I'm not 100% sure that the behavior of adding indices for background components is correct in the first place, since it seems inconsistent with what happens with no masks passed in.

# Type of change

# Has your PR been tested?

The CNMF run I that was failing for me before now succeeds. Since there wasn't a failing test before, I'm guessing this part of the code (under the condition of seeded CNMF) wasn't being tested, which could be changed.

### DIFF
--- a/caiman/source_extraction/cnmf/spatial.py
+++ b/caiman/source_extraction/cnmf/spatial.py
@@ -180,7 +180,7 @@ def update_spatial_components(Y, C=None, f=None, A_in=None, sn=None, dims=None,
         A_in = caiman.utils.stats.csc_column_remove(A_in, list(ff))
         C = np.delete(C, list(ff), 0)
         # update indices
-        ind_list = list(range(nr-np.size(ff)))
+        ind_list = list(range(nr+nb-np.size(ff)))
         for i in ff:
             ind_list.insert(i, 0)
         ind_list = np.array(ind_list, dtype=int)


pgunn:
We're likely to merge this after the next release; if you can test this the best you can with a variety of settings in the meantime that'd be great. Sorry for the delay.

ethanbb:
> However, I'm not 100% sure that the behavior of adding indices for background components is correct in the first place, since it seems inconsistent with what happens with no masks passed in.

I'm pretty sure now that this is correct and the code works as designed with this fix. It's pretty clearly intentional later in `update_spatial_components` that the background components are included in the regression step:

https://github.com/flatironinstitute/CaImAn/blob/4a22789176833c8fcd6daedf2841e723d9d673aa/caiman/source_extraction/cnmf/spatial.py#L171-L173

And this makes sense based on the problem statement in `regression_ipyparallel`, which fits each pixel's component weights including background components:

https://github.com/flatironinstitute/CaImAn/blob/4a22789176833c8fcd6daedf2841e723d9d673aa/caiman/source_extraction/cnmf/spatial.py#L285-L288

Furthermore, in the case where A_in is not boolean, the input background spatial components `b` are also included when computing the indices (if non-None), so this fix is probably sometimes necessary for this case as well. (This could probably be simplified to just adding all ones for the background components though, since I think each background component can have a nonzero weight on any pixel):

https://github.com/flatironinstitute/CaImAn/blob/4a22789176833c8fcd6daedf2841e723d9d673aa/caiman/source_extraction/cnmf/spatial.py#L1023-L1029

I usually don't supply `b_in` when I use a boolean `A_in`, but it looks like for non-boolean `A_in`, `b_in` is required or else it's assumed that none of the background components load onto any of the pixels. I believe `b` and `f` will come in as None unless provided by the user because the initialization step is skipped, and neither is initialized in the non-bool case. Overall it seems like handling of the background components in the non-bool case when only `A_in` is given is not well-tested/potentially broken and should probably work more like the bool case.

By the way, in #1539, @zyyk78 you state that this line in `computing_indicator` removes indices >= nr for the non-bool case:
```python
ind2_ = [iid_ if (np.size(iid_) > 0) and (np.min(iid_) < nr) else [] for iid_ in ind2_]
```

However, this isn't true. Note that the condition is `np.min(iid_) < nr` - the *minimum* index can't be one of the background components, which means the component can't be _all_ background, but it can include background, as it would be expected to if a reasonable `b_in` is provided.

zyyk78:
>  However, this isn't true. Note that the condition is np.min(iid_) < nr - the minimum index can't be one of the background components, which means the component can't be all background, but it can include background, as it would be expected to if a reasonable b_in is provided.     

To be honest, this ind2_  may include background components, and this may cause errors like it .   
   
But i don't understand why i didn't meet this while using without binary masks (maybe removing empty component is very rare in that case ?)

ethanbb:
> But i don't understand why i didn't meet this while using without binary masks (maybe removing empty component is very rare in that case ?)

I think because if you pass `A_in` but not `b_in`, it does this:

```python
 if b is None: 
     dist_indicator = determine_search_location( 
         A_in, dims, method=method, min_size=min_size, max_size=max_size, dist=dist, expandCore=expandCore, dview=dview) 
```

which just looks at the spatial components and doesn't try to add any indices for the background components. But as I said above, this doesn't seem correct to me.

pgunn:
FYI, I'll keep an eye on the discussion and most likely will merge this after Manuel's current PR lands, before the next release, unless the discussion moves towards a conclusion to fix it a different way.

ethanbb:
I think what I would like to do at some point is add tests for both the boolean and non-boolean seed cases, confirm that it currently fails for the boolean case, and see whether the results are also weird (or erroring) for the non-boolean case due to not starting with the background components included. I suspect the code for that case can be simplified to make it more like the boolean case, aside from binarizing the masks, but I want to see whether it's actually currently broken first.

pgunn:
@ethanbb I landed Manuel's changes; after the dust settles I'd be happy to take this back up

Issue #1417, opened 2024-10-29
Add dependencies to `pyproject.toml`


sneakers-the-rat:
Hello!

I see a previous issue on packaging re: adding to pypi - https://github.com/flatironinstitute/CaImAn/issues/509 - but this is sufficiently distinct, and the python packaging world has changed dramatically since 2019.

Currently the `pyproject.toml` file doesn't specify any dependencies and they are just listed in `requirements.txt`

Moving the `requirements.txt` to `pyproject.toml` would be essentially a no-effort change that allows the package to be installed via pip/any standards-compliant package manager, and if there are differences in the gpu requirements, those could be added as a `extra` or `optional` dependencies.

Doing that should also allow you to make a pypi-ready package for free, even if you just put source distributions up there and don't want to worry about building wheels for various platforms and versions, but most importantly it allows people to build downstream tools on top of CaImAn. Currently that's impossible unless you add all the requirements to downstream packages manually, and then that package is on the hook with maintaining a parallel set of deps. Similarly, not being on pypi prevents people building downstream tools except via conda, which is not really ideal for making a generalizable python tool.

Should be no additional maintenance burden (and in fact a lesser one where you can single-source the deps in a format that all python package tooling will support for the indefinite future).

I can PR if y'all are strapped for time, trying it locally and it works exactly as expected. :)

pgunn:
Right now we're mostly only supporting conda-forge based installs, although I see little harm in doing this. We mostly provide requirements.txt for people either building from source (which we mostly discourage) or for people running in colab. 

My understanding, based on the last time I read about these packaging changes, is that pyproject actually can work fine alongside requirements.txt - one doesn't need to pick one or the other and the tooling will generally use both if present, correctly. Is that no longer the case? I think we probably don't need to focus as much on legacy compatibility though, so this is probably a good thing (even though I'm not sure how high-priority it should be given that it's unclear to me if it actually affects anything).

I'll put this on the todo list.

sneakers-the-rat:
> building from source (which we mostly discourage) 

totally understand not wanting to take on another packaging modality due to maintenance burden, but hear me out! If you didn't want to upload to pypi, just adding the deps to `pyproject.toml` still has benefits without drawbacks, and if you did want to upload to pypi, doing so would also allow you to build working wheels so people wouldn't need to install from source :)

> pyproject actually can work fine alongside requirements.txt 

[...]

PR #1432, opened 2024-12-06
Move dependencies into pyproject

pgunn:
Addresses #1417 

Thanks to @sneakers-the-rat for prodding us on this.

I made jupyter the optional dependency here; we could also break the "dev" tools out, but the test functionality of caimanmanager depends on mypy and I'd rather not go through the fuss to make it optional (mypy is a fairly small and no-fuss dependency anyhow, so far).


pgunn:
Seems to pass at least preliminary testing using a bare conda environment and a pip-only install. It's not a very fast install route, but it seems to work.

pyproject.toml
```diff
@@ -11,6 +11,43 @@ authors = [
...
+  "tensorflow",
+  "tifffile",
+  "tqdm",
+  "yapf",
+  "zarr"
+]
+
+[project.optional-dependencies]
+jupyter = [
+  "jupyter",
```
kushalkolar, line 47: It doesn't make sense for jupyter to be optional but bokeh mandatory?

pgunn: CNMF and visualization both import bokeh; I think we need it to be present until some future cleanup changes that.

I don't think bokeh pulls jupyter in.

kushalkolar: Ah right. Though none of the bokeh viz will actually work without jupyter AFAIK.

Issue #1416, opened 2024-10-26
Propose removing old unmaintained skcuda code 


kushalkolar:
Should probably remove all the cuda stuff from motion correction because it just confuses people, example: #1048 . skcuda is no longer maintained, [last release was over 5 years ago](https://github.com/lebedov/scikit-cuda/tags), so it's very unlikely to work with any relatively recently CUDA version and GPU, and would be difficult to interface with any other software. Given these limitations I don't even have a way to test if this even works. If users want GPU accelerated NoRMCorre we should just redirect them to FIOLA or jnormcorre which do a great job.

https://github.com/flatironinstitute/CaImAn/blob/bb55800806f0898592d79dcc705a0b53ccd01ec3/caiman/motion_correction.py#L1629-L1636

pgunn:
It's not entirely unmaintained; the author seems to occasionally still be accepting patches, but no release in nine years isn't great.

I think this proposal seems reasonable; the other option would be to port it to some other framework (which would at least not require people to go out-of-codebase) but I'm not keen to volunteer for such work; unless someone else shows up keen to do that soon, removing these codepaths is probably a good idea.

pgunn:
Now in dev

PR #1433, opened 2024-12-06
Remove skcuda from motion_correction. Implements #1416

pgunn:


caiman/motion_correction.py
```diff
@@ -143,8 +135,8 @@ def __init__(self, fname, min_mov=None, dview=None, max_shifts=(6, 6), niter_rig
            nonneg_movie: boolean
                make the SAVED movie and template mostly nonnegative by removing min_mov from movie
 
-           use_cuda : bool, optional
-               Use skcuda.fft (if available). Default: False
+           use_cuda : boolean (DEPRECATED)
+               No longer used, may be removed in a future version
```
kushalkolar, line 139: Kinds confusing when I read it first, what about "cuda is no longer supported for motion correction, this kwarg will be removed in a future version. See the jnormcorre project for a GPU accelerated implementation."

caiman/motion_correction.py
```diff
@@ -69,15 +69,7 @@
...
-opencv = True
-
-try:
-    import pycuda.gpuarray as gpuarray
-    import pycuda.driver as cudadrv
-    import atexit
-    HAS_CUDA = True
-except ImportError:
-    HAS_CUDA = False
+opencv = True # FIXME How is the user meant to be able to interact with this?!
```
kushalkolar, line 72: Yea this is weird. Also skimage uses opencv under the hood for many things so I wonder what this actually does.

caiman/motion_correction.py
```diff
@@ -204,8 +196,8 @@ def __init__(self, fname, min_mov=None, dview=None, max_shifts=(6, 6), niter_rig
         self.var_name_hdf5 = var_name_hdf5
         self.is3D = bool(is3D)
         self.indices = indices
-        if self.use_cuda and not HAS_CUDA:
-            logger.debug("pycuda is unavailable. Falling back to default FFT.")
+        if self.use_cuda:
+            logger.warn("skcuda is no longer supported; using non-cuda algorithms")
```
kushalkolar, line 200: Add full deprecation warning text here too?

caiman/motion_correction.py
pgunn: I'm not sure about mentioning jnormcorre until/unless we have docs saying how to use it with the rest of the codebase.

The rest seems sensible.

caiman/motion_correction.py
pgunn: Will do

caiman/motion_correction.py
pgunn: This may be a remnant of the spyder era, when the line between library code and code the user might edit was very thin. It was fairly unstructured and messy.

Maybe before this is ready for commit I'll stare at this a lot more and maybe dig through the commit history - it may be better to clean this up than just add a note. Although maybe that's better done as a separate diff.

pgunn: The code is ancient, introduced in 2e7f1a2719f38cd2016038e27f398fa7d2f9a339 (December 2016) during the spyder era when the code was not yet really following library semantics.

We probably should remove the (now probably long-untested) opencv=False code path.

caiman/motion_correction.py
```diff
@@ -137,31 +124,37 @@ def __init__(self, fname, min_mov=None, dview=None, max_shifts=(6, 6), niter_rig
...
                apply shifts fast way (but smoothing results)
 
-           nonneg_movie: boolean
+           nonneg_movie: bool
                make the SAVED movie and template mostly nonnegative by removing min_mov from movie
 
-           use_cuda : bool, optional
-               Use skcuda.fft (if available). Default: False
+           gSig_filt: list
+               (UNDOCUMENTED)
```
kushalkolar, line 134:
```suggestion
           gSig_filt: list[int], default None
               if specified, indicates the shape of a Gaussian kernel used to perform high-pass spatial filtering of the video frames before performing motion correction. Useful for data with high background to emphasize landmarks. 
```

caiman/motion_correction.py
```diff
@@ -137,31 +124,37 @@ def __init__(self, fname, min_mov=None, dview=None, max_shifts=(6, 6), niter_rig
...
+           use_cuda : bool (DEPRECATED)
+               cuda is no longer supported for motion correction; this kwarg will be removed in a future version of Caiman
 
            border_nan : bool or string, optional
                Specifies how to deal with borders. (True, False, 'copy', 'min')
                TODO: make this just the bool, and make another variable called
                      border_strategy to hold the how
 
+           pw_rigid: bool, default: False
+               flag for performing motion correction when calling motion_correct
```
kushalkolar, line 145:
```suggestion
               flag for performing elastic motion correction when calling motion_correct
```

kushalkolar (changes requested):

PR #1434, opened 2024-12-11
Windows users: use pip and compile it

pgunn:
Not happy about this, but right now tensorflow doesn't have windows builds on conda-forge:

https://anaconda.org/conda-forge/tensorflow/files

and we can no longer use the defaults channel for caiman. This means binary conda packages of caiman are not going to happen on Windows until some things change, most likely a pair of things:
A) The upcoming switchover of our code from tensorflow to pytorch (but pytorch also lacks packaging on conda-forge on Windows!)
B) The upstream efforts to get pytorch's packaging issues on conda-forge on Windows fixed: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231

At that point we can revert these changes and resume packaging conda packages of caiman on Windows, but for now we'll instruct them to install a compiler and use pip. (a recent pyproject diff here at least makes this temporary measure slightly more robust probably).

(we are considering adding pip support as a long-term supported install route but haven't decided on it yet)

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -29,6 +29,7 @@ jobs:
           environment-file: environment-minimal.yml
           activate-environment: caiman
           conda-solver: libmamba
+          miniforge-version: latest
 
       - name: Install OS Dependencies
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


pgunn:
This unblocks us on releases, at least for OSX and Linux.

PR #1435, opened 2024-12-12
remove skcuda stuff from docs

kushalkolar:
w.r.t. changes in #1433 

btw this is the first time I'm seeing "search for the most recent opencv with openblas and switch to it using existing conda tooling", I have no idea how relevant that is O_O


pgunn:
I wonder if we should just remove readme-gpu; this is a step in the right direction though.

PR #1444, opened 2025-01-05
Fix and add more options for non-optical flow register_ROIs; add feature to interpolate shifts based on patch locations in tile_and_correct

ethanbb:
# Description

## The problem

This addresses an issue I raised on the slack where the non-optical flow path of `register_ROIs` seems to have a mistake. It works by finding piecewise-rigid shifts between the two template images using `tile_and_correct`. Then, the matrix of shifts for each patch should be interpolated into a shift map of the same size as the image, to be used for `cv2.remap`. However, these lines:

https://github.com/flatironinstitute/CaImAn/blob/bb55800806f0898592d79dcc705a0b53ccd01ec3/caiman/base/rois.py#L434-L435

actually just [repeat the entries of the matrix, in C-order, to fill in a new matrix of the requested shape](https://numpy.org/doc/1.26/reference/generated/numpy.resize.html), so the result will make no sense.

I used this test script: https://gist.github.com/ethanbb/12e6efa6ba31733f769c0b6c3f82ee98 to confirm that it gives weirdly warped results. Note that with the current code, setting `use_opt_flow` to False always gives an error, because the `tile_and_correct` call sets `shifts_opencv=True`, but this path does not return a value for `xy_grid`, which is then used on the next line. So, for testing, I temporarily changed `shifts_opencv` to False. Here are the results with the current code:

register_rois_test_result_orig_dev2

In this case, it seems like we should just stick with optical flow. However, one thing I've observed and which motivated this inquiry is that optical flow struggles with larger shifts. Here is what happens when I increase the amount of non-rigid offset from 2 to 6:

register_rois_test_result_orig_dev6

Clearly, neither method is satisfactory.

## Solving the general problem

Up-sampling shifts from patches does not only happen in `register_ROIs`. It occurs in multiple places in the motion correction code as well, but for this PR I just focused on `tile_and_correct` and `tile_and_correct_3d`. In these functions, resizing occurs for both the OpenCV (to full image size in order to use `cv2.remap`) and non-OpenCV (to an Nx denser grid, depending on `upsample_factor_grid`) methods. I realized when looking into this that it was not being done in the most rigorously correct way; it is implemented as a standard image resize, but the patch centers at the borders do not coincide with the borders of the new image, and these centers may not even be evenly spaced (at the lower and right edges).

This imprecision may not usually be a big deal for large images, and may be necessary in order to correct whole movies efficiently. However, I thought that at least for the case of aligning just 2 images, I may as well add a way to up-sample more precisely by interpolating using the exact location of each patch. Thus, I added `get_patch_centers` and `interpolate_shifts` to interpolate shifts to/from specific pixel locations using scipy (I don't think OpenCV has an easy way of doing this). I also added the flag `interp_shifts_precisely` for the two tile_and_correct functions to enable this method. I haven't tested it on a full motion correction run yet, but it could potentially be promoted to a new parameter in the future. 

Finally, I used the same functions to do the interpolation correctly within `register_ROIs`. Using the new `get_patch_centers` function, the `xy_grid` output is unnecessary, so it works with either setting of `shifts_opencv`. I also set `interp_shifts_precisely` to True here, and it seems to be working fine.

Results with `nr_offset = 2`:
register_rois_test_result_fixed_dev2

And with `nr_offset = 6`:
register_rois_test_result_fixed_dev6

## Other changes to register_ROIs

- I added an optional argument `align_options` to allow overriding default motion correction params (for `tile_and_correct` method only). This is necessary, among other things, to increase `max_deviation_rigid` from 2 in order to deal with cases like the second example above.
- I added a clause to deal with the rigid case (if setting `max_deviation_rigid = 0`)
- For the optical flow branch, I noticed that the images are being multiplied by 255 and converted to uint8 twice in a row, causing lots of overflow. I assumed this was a bug, so I removed the second one.

# Type of change

# Has your PR been tested?

Aside from the [test script]( https://gist.github.com/ethanbb/12e6efa6ba31733f769c0b6c3f82ee98) mentioned above, `caimanmanager test` passes. I'm trying to run demotest, having some issues with it but I'll update if I can get it working.

### DIFF
--- a/caiman/base/rois.py
+++ b/caiman/base/rois.py
@@ -430,12 +430,12 @@ def register_ROIs(A1,
                 "max_shifts": (10, 10),
                 "shifts_opencv": True,
                 "upsample_factor_grid": 4,
-                "interp_shifts_precisely": True,
+                "shifts_interpolate": True,
                 "max_deviation_rigid": 2
                 # any other argument to tile_and_correct can also be used in align_options
             }
 
-            if align_options is not None:
+            if align_options:
                 # override defaults with input options
                 align_defaults.update(align_options)
             align_options = align_defaults
--- a/caiman/base/rois.py
+++ b/caiman/base/rois.py
@@ -20,7 +20,7 @@ import time
 from typing import Any, Optional
 import zipfile
 
-from caiman.motion_correction import tile_and_correct
+from caiman.motion_correction import tile_and_correct, get_patch_centers, interpolate_shifts
 
 try:
     cv2.setNumThreads(0)
@@ -414,25 +414,31 @@ def register_ROIs(A1,
         if use_opt_flow:
             template1_norm = np.uint8(template1 * (template1 > 0) * 255)
             template2_norm = np.uint8(template2 * (template2 > 0) * 255)
-            flow = cv2.calcOpticalFlowFarneback(np.uint8(template1_norm * 255), np.uint8(template2_norm * 255), None,
+            flow = cv2.calcOpticalFlowFarneback(template1_norm, template2_norm, None,
                                                 0.5, 3, 128, 3, 7, 1.5, 0)
             x_remap = (flow[:, :, 0] + x_grid).astype(np.float32)
             y_remap = (flow[:, :, 1] + y_grid).astype(np.float32)
 
         else:
-            template2, shifts, _, xy_grid = tile_and_correct(template2,
-                                                             template1 - template1.min(),
-                                                             [int(dims[0] / 4), int(dims[1] / 4)], [16, 16], [10, 10],
-                                                             add_to_movie=template2.min(),
-                                                             shifts_opencv=True)
+            strides = (int(dims[0] / 4), int(dims[1] / 4))
+            overlaps = (16, 16)
+            max_shifts = (10, 10)
+            shifts_opencv = True
+            upsample_factor_grid = 4
+            template2, shifts, _, _ = tile_and_correct(
+                template2, template1 - template1.min(), strides=strides, overlaps=overlaps,
+                max_shifts=max_shifts, add_to_movie=template2.min(),
+                upsample_factor_grid=upsample_factor_grid, shifts_opencv=shifts_opencv, interp_shifts_precisely=True)
 
-            dims_grid = tuple(np.max(np.stack(xy_grid, axis=0), axis=0) - np.min(np.stack(xy_grid, axis=0), axis=0) + 1)
+            patch_centers = get_patch_centers(dims, overlaps=overlaps, strides=strides,
+                                              shifts_opencv=shifts_opencv, upsample_factor_grid=upsample_factor_grid)
+            patch_grid = tuple(len(centers) for centers in patch_centers)
             _sh_ = np.stack(shifts, axis=0)
-            shifts_x = np.reshape(_sh_[:, 1], dims_grid, order='C').astype(np.float32)
-            shifts_y = np.reshape(_sh_[:, 0], dims_grid, order='C').astype(np.float32)
+            shifts_x = np.reshape(_sh_[:, 1], patch_grid, order='C').astype(np.float32)
+            shifts_y = np.reshape(_sh_[:, 0], patch_grid, order='C').astype(np.float32)
 
-            x_remap = (-np.resize(shifts_x, dims) + x_grid).astype(np.float32)
-            y_remap = (-np.resize(shifts_y, dims) + y_grid).astype(np.float32)
+            x_remap = interpolate_shifts(-shifts_x, patch_centers, tuple(range(d) for d in dims))
+            y_remap = interpolate_shifts(-shifts_y, patch_centers, tuple(range(d) for d in dims))
 
         A_2t = np.reshape(A2, dims + (-1,), order='F').transpose(2, 0, 1)
         A2 = np.stack([cv2.remap(img.astype(np.float32), x_remap, y_remap, cv2.INTER_NEAREST) for img in A_2t], axis=0)
--- a/caiman/base/rois.py
+++ b/caiman/base/rois.py
@@ -312,14 +312,14 @@ def register_ROIs(A1,
                   D=None,
                   max_thr=0,
                   use_opt_flow=True,
-                  align_options: Optional[dict] = None,
                   thresh_cost=.7,
                   max_dist=10,
                   enclosed_thr=None,
                   print_assignment=False,
                   plot_results=False,
                   Cn=None,
-                  cmap='viridis'):
+                  cmap='viridis',
+                  align_options: Optional[dict] = None):
     """
     Register ROIs across different sessions using an intersection over union 
     metric and the Hungarian algorithm for optimal matching
@@ -352,9 +352,6 @@ def register_ROIs(A1,
         use_opt_flow: bool
             use dense optical flow to align templates
 
-        align_options: Optional[dict]
-            mcorr options to override defaults when align_flag is True and use_opt_flow is False
-
         thresh_cost: scalar
             maximum distance considered
 
@@ -376,6 +373,9 @@ def register_ROIs(A1,
 
         cmap: string
             colormap for background image
+        
+        align_options: Optional[dict]
+            mcorr options to override defaults when align_flag is True and use_opt_flow is False
 
     Returns:
         matched_ROIs1: list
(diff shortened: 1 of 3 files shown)


caiman/motion_correction.py
```diff
@@ -1799,15 +1800,69 @@ def apply_shifts_dft(src_freq, shifts, diffphase, is_freq=True, border_nan=True)
...
 
     Args: 
-        img:ndarray 2D
-            image that needs to be slices
+        img: ndarray 2D
+            image that needs to be sliced
 
         overlaps: tuple
-            dimension of the patch
+            overlap of patches (except possibly last one) in each dimension
```
pgunn, line 1865: The "except" part probably could use some explanation

pgunn (commented):
Very nice work; do you mind cherry-picking the CI diffs from dev into this so tests can run?

ethanbb:
> Very nice work; do you mind cherry-picking the CI diffs from dev into this so tests can run?

Thanks! Can you be more specific - this should already be applied on top of dev. Are the CI changes in another branch?

pgunn:
> > Very nice work; do you mind cherry-picking the CI diffs from dev into this so tests can run?
> 
> Thanks! Can you be more specific - this should already be applied on top of dev. Are the CI changes in another branch?

They should be in dev now; the problem is github's test infrastructure runs on your branch, not on what the dev branch would look like with your changes, so your branch being slightly out of date means the fixes are not effective.

(rebasing would also work)

ethanbb:
Sorry would you mind telling me the specific commits you're talking about? As far as I can tell it's not behind flatironinstitute/dev.

pgunn:
Are you ready for this to be merged now?

ethanbb:
I think it's up to you - I have been using it fine on my own branch. I wasn't able to get the demo tests working locally. I think it might be worth considering whether that interp_shifts_precisely option should be made into a real parameter - I would like to try using it to motion correct a movie at some point and see how it differs in terms of runtime and results.

pgunn:
Let me see if the CI stuff is quick to fix.

pgunn:
Passes CI now, which is good; you still sound uncertain on whether you want this style of parameterisation or something else; make a call on that. We could merge this now, or if you want to rework some of the approach first that's also fine.

ethanbb:
OK, I tried motion correcting a long recording using interp_shifts_precisely. It made very little difference in the result, but it also didn't take much longer than normal, so I decided it's fine to add it as a new parameter (false by default, though).

ethanbb:
Not done yet, but when going through the code, I noticed a couple of things that seem to be bugs:
- When applying shifts to a 2D image in `tile_and_correct` using opencv remap, the `border_nan` option isn't taken into account (even though it is used when doing the same operation in `apply_shifts_movie`)
- In `apply_shifts_movie`, for the 3D case, the shifts aren't un-negated as they are for the 2D case (as far as I can tell there should be no difference)

ethanbb:
OK, now I'm done with this.

caiman/source_extraction/cnmf/params.py
```diff
@@ -858,6 +861,7 @@ def __init__(self, fnames=None, dims=None, dxy=(1, 1),
         self.motion = {
             'border_nan': 'copy',               # flag for allowing NaN in the boundaries
             'gSig_filt': None,                  # size of kernel for high pass spatial filtering in 1p data
+            'interp_shifts_precisely': False,   # interpolate shifts based on patch locations instead of reizing
```
pgunn, line 864: I wonder if precise_shifts or something else that's a little more concise would be better.

caiman/base/rois.py
```diff
@@ -414,25 +418,50 @@ def register_ROIs(A1,
...
+                "overlaps": (16, 16),
+                "max_shifts": (10, 10),
+                "shifts_opencv": True,
+                "upsample_factor_grid": 4,
+                "interp_shifts_precisely": True,
+                "max_deviation_rigid": 2
+                # any other argument to tile_and_correct can also be used in align_options
+            }
+
+            if align_options is not None:
```
pgunn, line 438: Maybe prettier as `if align_options:`

caiman/source_extraction/cnmf/params.py
ethanbb: Definitely could change the name...I think that just "precise_shifts" might be too vague, since it's not about the values of the shifts themselves but just how they're interpolated?

caiman/motion_correction.py
```diff
@@ -2464,6 +2497,117 @@ def tile_and_correct_3d(img:np.ndarray, template:np.ndarray, strides:tuple, over
...
+        shifts_z = interpolate_shifts(shift_img_z, patch_centers, coords_new).astype(np.float32)
+    else:
+        shifts_y = resize_sk(shift_img_y.astype(np.float32), dims)
+        shifts_x = resize_sk(shift_img_x.astype(np.float32), dims)
+        shifts_z = resize_sk(shift_img_z.astype(np.float32), dims)
+    
+    shift_map = np.stack((shifts_y + y_grid, shifts_x + x_grid, shifts_z + z_grid), axis=0)
+    
+    # apply to image
+    if border_nan is False:
```
pgunn, line 2594: if not border_nan:

caiman/motion_correction.py
```diff
@@ -2464,6 +2497,117 @@ def tile_and_correct_3d(img:np.ndarray, template:np.ndarray, strides:tuple, over
...
+        mode = 'constant'
+        value = 0.0
+    elif border_nan is True:
```
pgunn, line 2597: elif border_nan:

caiman/base/rois.py
ethanbb: Sure

caiman/motion_correction.py
ethanbb: This isn't good because here, True is treated differently from other truthy values (the strings 'min' and 'copy').

caiman/source_extraction/cnmf/params.py
pgunn: I'm not set on that name, but "interp" kind of bugs me and the whole thing seems a bit long.

caiman/motion_correction.py
pgunn: Ah, fair enough.

pgunn: I'm worried we may have made mistakes in allowable values that we need to think about this, but it's probably too late to fix that now without a lot more work that's out of context here.

caiman/source_extraction/cnmf/params.py
ethanbb: How about just "interpolate_shifts"? precisely is redundant, it's either interpolating or resizing

caiman/motion_correction.py
ethanbb: technically this would work, but given the fact that it's a weird type I feel like it's better to be more explicit.

pgunn: Fair enough.

caiman/source_extraction/cnmf/params.py
pgunn: Sure, that works.

ethanbb: Oh, wait, that actually conflicts with the name of one of the functions I added. It's probably better as the name of a function. What if it was a string parameter instead, either "resize" or "interpolate"? Maybe call it "upsampling_method" or can you think of something better?

pgunn: Do we expect there to be other reasonable ways to do this we might want to implement someday? If not, a boolean should be fine.

If we want to make it string, upsample_method isn't bad, or maybe just upsample? shift_resize?

ethanbb: > Do we expect there to be other reasonable ways to do this we might want to implement someday? If not, a boolean should be fine.

Yeah, I don't know, probably not. 

How about "pw_interpolate"?

pgunn: Heh, we're definitely proving that "naming things is hard" adage.

I just noticed the text says "reizing" rather than resizing - that needs a fix.

For the name, maybe shifts_interpolate is the most clear. "pw" feels unclear to me.

ethanbb: Works for me!

ethanbb:
Should be all set now.

PR #1452, opened 2025-01-18
Fix incorrect framerate when loading bidirectional SBX files

ethanbb:
# Description

I realized that the frame rate when loading a bidirectional SBX file was half of what it should be. This is because in `sbx_meta_data`, I assumed that `info['recordsPerBuffer']` would be half for bidirectional files, so dividing by it when computing the frame rate would give 2X the rate for these files, as it should be. But I must have forgotten I was doing this, because in `sbx_shape` (which is called first) I modify `info[recordsPerBuffer]` to match the X size. This is unnecessary since I already have the X size in `info['sz'][0]`.

Along the way, I also made a few miscellaneous changes to eliminate type errors according to pyright (some important for documentation, such as the requirement that subindices be sequences, not just iterables).

# Type of change

# Has your PR been tested?

Yes, I added a test to test_sbx.py that loads a new bidirectional test movie and ensures the sample rate is correct. I also added checks for the sample rates of the other test movies.


ethanbb:
My bad, I added a return value check that was in my branch thinking it was just for safety, but I didn't realize that the return value was expected to be None unless I passed a specific option (returnoffset for tifffile.imwrite). Will undo that.

Issue #1454, opened 2025-01-21
volpy gui doesn't work with Qt6 (or Qt5, except on some platforms)


pgunn:
The volpy gui doesn't seem to work with qt6, which is delivered by default with most environment solves; it has a dependency on PyQt5.

It's unclear how hard it would be to update `caiman/source_extraction/volpy/volpy_gui.py` to accept qt6.

@caichangjia giving you a heads-up on this; I'll look into it too but you're much more familiar with that code.

pgunn:
Hm. Merely installing pyqt into the environment also doesn't seem to produce a working app.

```
scclin022:~/caiman_data$ python /mnt/home/pgunn/src/caiman_dev/caiman/source_extraction/volpy/volpy_gui.py
2025-01-21 14:13:25.410789: I tensorflow/core/util/port.cc:113] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2025-01-21 14:13:25.447661: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:9261] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-21 14:13:25.447697: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:607] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-21 14:13:25.448693: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1515] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
2025-01-21 14:13:25.454339: I tensorflow/core/platform/cpu_feature_guard.cc:182] This TensorFlow binary is optimized to use available CPU instructions in performance-critical operations.
To enable the following instructions: SSE4.1 SSE4.2 AVX AVX2 AVX512F AVX512_VNNI FMA, in other operations, rebuild TensorFlow with the appropriate compiler flags.
Qt: Session management error: None of the authentication protocols specified are supported
QWidget: Must construct a QApplication before a QWidget
Aborted (core dumped)
```

pgunn:
Ah, that warning or error can be solved by unsetting the SESSION_MANAGER env var. Still core dumping on me; unclear why.

pgunn:
[...]

PR #1455, opened 2025-01-22
Addresses #1454 using pyside6

pgunn:
Upstream changes in pyqtgraph late last year broke the volpy gui because mixing pyqtgraph and PyQt5 became impossible (pyqtgraph routed itself through pyside6 if it's present, causing the app to mix PyQt5 and QT6 objects).

This diff commits volpy to using QT6 through the pyside6 library, avoiding possibilities of mixing them.

### DIFF
--- a/caiman/source_extraction/volpy/volpy_gui.py
+++ b/caiman/source_extraction/volpy/volpy_gui.py
@@ -15,18 +15,19 @@ import pyqtgraph as pg
 from pyqtgraph import FileDialog
 from pyqtgraph.Qt import QtGui
 from pyqtgraph.parametertree import Parameter, ParameterTree
-import PyQt5
-from PyQt5 import QtWidgets
-from PyQt5.QtWidgets import QApplication, QShortcut
+import PySide6
+from PySide6 import QtWidgets
+from PySide6.QtWidgets import QApplication
+from PySide6.QtGui import QShortcut
 import random
 from skimage.draw import polygon
 import sys
 
 import caiman as cm
 from caiman.external.cell_magic_wand import cell_magic_wand_single_point
-    
+
 os.environ["QT_QPA_PLATFORM_PLUGIN_PATH"] = os.fspath(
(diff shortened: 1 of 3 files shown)


pgunn:
Will need to change the feedstock to add the pyside6 dependency too. May want to expose volpy_gui as a package binary?

caiman/source_extraction/volpy/volpy_gui.py
```diff
@@ -15,18 +22,15 @@
...
 from skimage.draw import polygon
 import sys
 
 import caiman as cm
 from caiman.external.cell_magic_wand import cell_magic_wand_single_point
-    
+
 os.environ["QT_QPA_PLATFORM_PLUGIN_PATH"] = os.fspath(
-    Path(PyQt5.__file__).resolve().parent / "Qt5" / "plugins")
+    Path(PySide6.__file__).resolve().parent / "Qt5" / "plugins")
```
kushalkolar, line 33: I wonder if this would still be valid with `Qt5` in the path when using pyside6?

environment.yml
```diff
@@ -27,6 +27,7 @@ dependencies:
 - psutil
 - pynwb
 - pyqtgraph
+- pyside6
```
kushalkolar, line 30: I wonder if including this in the deps could create issues with opencv, last time I remember it wants to have its own qt installed. That's why I always use `opencv-python-headless` but I'm not sure if opencv windowed stuff still works but just using the existing qt then.

pgunn: Looks like it pulls in qt6-main, so it's at least a closer version; more likely than sticking with qt5-based distros but something to watch out for.

Unfortunately, right now I think some of the code uses opencv's ability to launch a qt-based GUI, I think. I'd need to check to find out where but I'm pretty sure we do it.

pgunn:
After a lot of testing, I don't think the PyQt6 and PySide6 coexisting is something we're going to regularly see users face (nothing pulls this in automatically), and this fixes a real issue (pyqt which is QT5 no longer being well supported by other packages we use). I think this approach is the right way forward.

PR #1457, opened 2025-01-29
Numpy 2.0 compatibility: change all np.Inf to np.inf

fdeguire03:
# Description

NumPy 2.0 removed the `np.Inf` constant. To ensure compatibility with newer versions of NumPy, all instances of `np.Inf` have been replaced with `np.inf`. This change is fully backward-compatible, as `np.inf` has always been defined in NumPy.

# Type of change

# Has your PR been tested?

Verified that `np.inf` works across multiple NumPy versions.

### DIFF
--- a/caiman/base/movies.py
+++ b/caiman/base/movies.py
@@ -671,7 +671,7 @@ class movie(caiman.base.timeseries.timeseries):
         T, h, w = self.shape
         Y = np.reshape(self, (T, h * w))
         Y = Y - np.percentile(Y, 1)
-        Y = np.clip(Y, 0, np.Inf)
+        Y = np.clip(Y, 0, np.inf)
         estimator = sklearn.decomposition.NMF(n_components=n_components, init=init, tol=tol, **kwargs)
         time_components = estimator.fit_transform(Y)
         components_ = estimator.components_
--- a/caiman/source_extraction/cnmf/cnmf.py
+++ b/caiman/source_extraction/cnmf/cnmf.py
@@ -594,7 +594,7 @@ class CNMF(object):
                 if self.params.get('patch', 'nb_patch') > 0:
 
                     while len(self.estimates.merged_ROIs) > 0:
-                        self.merge_comps(Yr, mx=np.Inf, fast_merge=True)
+                        self.merge_comps(Yr, mx=np.inf, fast_merge=True)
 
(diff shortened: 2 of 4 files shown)


kushalkolar:
just wanted to double check and `Inf` is an alias for `inf`: https://numpy.org/doc/1.24/reference/constants.html#numpy.Inf

caiman/base/movies.py
```diff
@@ -671,7 +671,7 @@ def NonnegativeMatrixFactorization(self,
         T, h, w = self.shape
         Y = np.reshape(self, (T, h * w))
         Y = Y - np.percentile(Y, 1)
-        Y = np.clip(Y, 0, np.Inf)
+        Y = np.clip(Y, 0, np.inf)
```
kushalkolar, line 674: @pgunn I'm almost certain that the method this is in is actually no longer used for anyting in caiman. Lots of methods on `movie` from back when they were trying to implement basically the entirety of caiman in the movie class

kushalkolar (approved):

caiman/base/movies.py
pgunn: There are no calls to it in the repo; you're right.

After I merge this I'll go through the movie class and look for more things we can zap.

pgunn:
@fdeguire03 This isn't actually ever instance of np.inf in the codebase; it's a step in the right direction, but would you mind getting the rest too?

```
scclin022:~/src/caiman_dev$ grep -r np.inf . | grep -v build | cut -d: -f1 | uniq
./caiman/source_extraction/cnmf/initialization.py
./caiman/source_extraction/cnmf/oasis.cpp
./caiman/source_extraction/cnmf/params.py
./caiman/source_extraction/cnmf/cnmf.py
./caiman/source_extraction/cnmf/spatial.py
./caiman/source_extraction/cnmf/deconvolution.py
./caiman/source_extraction/cnmf/online_cnmf.py
./caiman/source_extraction/cnmf/oasis.pyx
./caiman/summary_images.py
./caiman/utils/stats.py
./caiman/utils/utils.py
./caiman/motion_correction.py
./caiman/base/movies.py
```

pgunn:
Oops, sorry, wrong grep there.

pgunn:
Merged. Thanks for the PR. Sorry for the confusion.

Issue #1462, opened 2025-02-07
Tracking: Fix calling semantics


pgunn:
Right now, the codebase has some methods that modify their object inline and then return a reference to their object. While this offers slight semantic usefulness in allowing for chaining of many operations in one line, it's also confusing because people might reasonably assume that these methods leave their object alone and return some kind of transformed copy (which is the more usual semantics).

This tracking issue is to find and remove this trait across the entire codebase (or, alternatively, to determine if there's some reason we need to keep this practice). In cases where it modifies itself inline and then returns self, we should just not return anything.

Methods of the CNMF class are what drew my attention to the issue (many years ago). There may be some other instances.

This will potentially break user code.

pgunn:
CNMF.refit should not be changed, because it returns a distinct object.
CNMF.fit_file should not be changed

CNMF.fit should be changed, because it modifies its object inline
CNMF.compute_residuals should be changed
CNMF.deconvolve should be changed
CNMF.HALS4traces should be changed
CNMF.HALS4footprints should be changed
CNMF.update_temporal should be changed
CNMF.update_spatial should be changed (also has some weird stuff in it)
CNMF.merge_comps should be changed
CNMF.initialize should be changed

pgunn:
Looks like a lot of the codebase does this; a "grep -r 'return self'" turns up matches all over the place. Starting with CNMF still seems sensible.

pgunn:
I think we've mostly fixed these; closing the ticket for now

PR #1464, opened 2025-02-07
Fix calling conventions

pgunn:
In pursuit of #1462, this series of diffs will reform the calling conventions in Caiman to not have methods return self, and to have calling/sample code in our codebase not rely on that.

This will hopefully make it more clear that when a method returns an object, it's a distinct object, and make our code more conventional. There's a lot of code to change; I don't know if I'll get everything in a single PR.

This will be a breaking change. There's also a chance that this may be a bad idea and we'll back off on this change, either just for now or permanently.

(this PR will change its title before it will be ready to merge)


pgunn:
In interest of not keeping a PR open until it's impossible to merge, I think it's probably best to consider this ready for merge now, with future work left to subsequent diffs.

PR #1474, opened 2025-02-26
Add versions to pyproject.toml

ethanbb:
# Description

Add versions to pyproject.toml to match environment.yml, so that `pip install .` works (I know it's not the preferred method, but may as well, and it's part of the testing for mesmerize-core).

Also I added another optional dependency group "dev" for testing.

# Type of change

# Has your PR been tested?

I tested on Linux.

- ```caimanmanager test``` works after doing `pip install ".[dev]"`
- Package imports successfully after doing `pip install .`
- Jupyterlab opens and caiman imports after doing `pip install ".[jupyter]"`


pgunn:
At least right now I'm not sure managing a separate set of test dependencies is a good idea; I'd like users to always be able to run "test" to verify their install works; it's not just a CI tool.

I'm open to the broader idea though, and recently I've been adjusting the docs to indicate a greater support for pip installs.

ethanbb:
OK, makes sense, I'll move the test dependencies into the main list.

pgunn:
Will merge when this goes green

kushalkolar:
Is caiman really not updated yet for numpy v2?!? numpy v2 brought some improvements in numerical stability, the changes required would be small if it's not already been udpated.

pgunn:
I think all that needs to happen is to flip the dependencies and put it through a run of testing. It's probably about time to do so.

ethanbb:
Current failures are mostly due to using `np.Inf` (capitalized) which was removed - maybe just in one place. See e.g. https://github.com/nel-lab/mesmerize-core/actions/runs/13558353507/job/37896896907?pr=325

Issue #1476, opened 2025-03-03
Speedup achieved by altering NMF parameters during initialization


fdeguire03:
The end of the `GreedyROI` function in `initialization.py` uses `sklearn.decomposition.NMF` to estimate the background component in each patch:

```
model = NMF(n_components=nb, init='nndsvdar')
b_in = model.fit_transform(np.maximum(res, 0)).astype(np.float32)
f_in = model.components_.astype(np.float32)
```

I have found that the parameters `init=random` and `max_iter=25` greatly improve the runtime (often by 2-3x) for this method without any detriment to quality (both methods give roughly the same reconstruction error in `model.reconstruction_err_`). Additionally, since `initialize_components` function immediately refines the results of `GreedyROI` using HALS, so using fewer iterations should not be an issue.

Would the maintainers consider switching to these `NMF` parameters by default or at least making them configurable via `params.py`?

pgunn:
We've exposed more parameters in the past when people requested them; I don't see any reason we couldn't do that here.

pgunn:
@kushalkolar I'm wondering if you have any thoughts on changing the defaults

kushalkolar:
Sure expose the params but I wouldn't change the defaults since that can mess with other people's known params.

pgunn:
@fdeguire03 I'll try to get these exposed by the next release; likely it'll be in dev within the week.

kushalkolar:
So the reason it's faster is because a random init will of course be faster than a non-negative SVD. However a nnsvd is generally more likely to converge than any random matrix.

fdeguire03:
Of course the random init will be faster than the SVD, so the question is whether the quality of the resulting factorization using the SVD initialization is good enough relative to the random initialization to justify the extra time. In my tests, I found that the random initialization (usually after only a few iterations) produced a factorization with a reconstruction error (measured by the `model._reconstruction_err` attribute) that was within about `1e-5` of the SVD-seeded factorization. Since the reconstruction errors were on the order of hundreds, this difference was negligible. Perhaps the random initialization just happened to work well on my data, but the speedup was significant enough (on my machine) that it may be worth investigating how much improvement the SVD initialization really provides on different datasets.

PR #1478, opened 2025-03-04
CNMF.initialization.greedyROI() - Add a limited mechanism to specify params in calls to sklearn's NMF

pgunn:
This adds a mechanism to greedyROI() to allow overriding of calls to sklearn's NMF(), passing in overrides as a dict.

(it also, as usual, includes some minor cleanups I spotted while looking at the code).

Will be WIP until the mechanism is plumbed up to CNMFParams, which will go through initialization.py:initialize_components() as well as utilities.py:manually_refine_components() although we may decide not to do it for the latter one maybe.

I went back-and-forth over whether to go with a dict or pass in params individually, but I think grouping them at least at this level is a little closer to where we eventually want to be - passing a structured deep dictionary parameters object around rather than having dozens of arguments everywhere and unwieldy gigantic signatures to all functions/methods.


pgunn:
This addresses #1476

pgunn:
Seems to work now, and it also fixes a few function comment issues along the way

PR #1502, opened 2025-05-21
Convert all uses of np.matrix to np.ndarray, handling API diffs

pgunn:
np.matrix is deprecated, and with changes made in numpy sometime back, it no longer offers a compelling use over np.ndarray. We now get deprecation warnings when using it; fortunately we don't pass np.matrix around that freely in the codebase so this hopefully is all we need to do.

### DIFF
--- a/caiman/source_extraction/cnmf/temporal.py
+++ b/caiman/source_extraction/cnmf/temporal.py
@@ -34,9 +34,9 @@ def make_G_matrix(T, g):
     if isinstance(g, np.ndarray):
         if len(g) == 1 and g < 0:
             g = 0
-        gs = np.matrix(np.hstack((1, -(g[:]).T)))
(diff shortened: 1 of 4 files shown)


pgunn:
Running into one of the remaining parts of API incompatibility between np.matrix and np.array ; np.matrix insists on having two dimensions and will automatically promote, while np.array is flexible and will take whatever shape it lands on.

So this refactor will take a bit more thought.

pgunn:
I worry about whether test coverage is adequate for this diff. Will need to do a fuller set of manual tests to see if I can find any problems. At aba262d, there is no more usage of np.matrix in the codebase, and pytest is actually looking reasonably happy.

(the code is unfortunately less clear, and I think it actually is better for code clarity to have something like np.matrix, but I'm not a numpy developer)

PR #1510, opened 2025-06-28
Fixed demo bug involving worker processes

tomjhou:
# Description

Two of the demos don't run because they spawn worker processes that re-import the main code, throwing errors and/or creating an infinite regress of additional processes. I fixed it it by adding the following check before running main():

```
if __name__ == "__main__":
    main()
```

# Type of change

# Has your PR been tested?

Yes the fix has been tested. Thank you for your consideration.


katonage:
i encountered the same issue

ethanbb:
Thank you, but this is a duplicate of #1565.

PR #1538, opened 2025-10-27
Take indices into account when applying piecewise shifts

ethanbb:
# Description

This fixes what I believe is an oversight in `apply_shifts_movie`. My assumption is that `apply_shifts_movie` is meant to be applied to a movie with the same spatial dimensions as the original movie mcorr was run on (e.g., corresponding functional imaging channel). Thus, if `indices` were specified to restrict motion correction to a certain field of view (or downsample by setting step to >1), the same indices should be used when applying piecewise shifts to put each shift in the right place. However, that is not currently done.

Since #1444 adds the `shifts_interpolate` method which interpolates and extrapolates (repeating edge) from the specific X/Y location of each piecewise shift, I just used that here, adjusting the patch centers to account for the indices. If any of the indices are not `slice(None)`, I took the simplest approach and forced the shifts_interpolate method.

# Type of change

# Has your PR been tested?

Unfortunately no, I don't think `apply_shifts_movie` is currently tested. I don't have time to add a test at this moment, but I might be able to add one later before merging.


pgunn:
This seems pretty reasonable, although I haven't really thought about mixing shifts and indices together (these felt like edge cases, haha). If you have a use for it and feel it's probably right, particularly if you add a test, I'm happy to merge this.

ethanbb:
> If you have a use for it 

Yeah, I am using indices to crop out features in the FOV that are not affected by brain motion (e.g., artifacts generated by the microscope always in the same place) just for motion correction, and then re-apply the shifts to the whole movie. I think this is useful functionality that should be supported...up to now I was only doing this with rigid shifts, but I recently realized that it should do the right thing for piecewise motion correction as well.

We can leave it for now and I can add a test when I have some time.

pgunn:
By leave it, do you mean do a merge of it as-is, or leave it open as a PR? I'm open to either.

ethanbb:
Oh, I meant leave open as a PR. Or hopefully I can at least get to the point of trying the code first. You know what, let me make it a draft PR until then, I don't want to merge code that hasn't been tested at all.

pgunn:
Ah, I may have misunderstood; I thought this was something you were already using but you haven't written discrete tests for it yet.

Sure, we can leave it as a draft.

ethanbb:
I'm noticing that the shifts interpolation when applying shifts is extremely slow, probably because I'm using cubic interpolation - cubic is also used in `cv2.resize` within tile_and_correct, but not in `apply_shifts_movie`. I'm not sure how intentional that is - I think it will be slow in either case, and I currently have `shifts_interpolate` turned off for normal motion correction because it's slow and doesn't make much difference. However, when applying shifts from a smaller FOV, it's pretty important to use the interpolation method, and using linear interpolation in that case is consistent with the existing method anyway. So I'm going to change the calls in `apply_shifts_movie` to use linear.

ethanbb:
Marking ready since I've tested it on my own data and it seems to be doing the right thing (& is faster now)

ethanbb:
Dang I just realized that when there is a step != 1 in any of the indices slices, the shifts themselves for the corresponding dimension should also be multiplied by that number to get the same effect. And arguably this is true for rigid shifts as well, although that may be more surprising (and breaking) so worth considering. I know this is an edge case (which I'm not using, which is why I didn't think of it), but that should be correct, right? @pgunn 

Not top priority but it would be ideal to get some tests for applying shifts with indices. If I have some extra time I can at least do a quick check for this case.

pgunn:
@ethanbb I think you're right, but I'd probably need to run the code with those parameters to refamiliarise myself to be sure. Which demonstrates your point - good testing is important.

I think some people use shifts (maybe very few), but theere've been times when we've found half-implemented ideas (that misbehaved) or code that never worked when fully testing some of the parameters. I've tried to reduce such things, but it never went as far as it should've. If the project had ever had a sustained dev team of, say, 5 people at 100% of their time, it would've been great to have one of them go through every function, figure out if it's used at all, test its parameters, make sure there's adequate test coverage, and put it in a system diagram.

Issue #1354, opened 2024-05-18
Tracking issue: Caiman and numpy 2.0


pgunn:
numpy 2.0 has release candidates, so a full release is just around the corner.

Right now caiman pins numpy at <2.0.

There's no need to try to test/fix our code before the 2.0 release, but it should be on our radar to adjust our code for 2.0 in the mid-term. Depending on the issues that come up we may:
A) Be able to get along with both versions, or
B) Need to treat this as a gate, doing a minimum version pin after the adjustments

I see both as acceptable, although the former is preferable because we'd be less all-elbows with other packages and user environments.

kushalkolar:
numpy v2 just released a few hours ago and is already causing problems elsewhere :melting_face: , need to wait for other packages (such as skimage) to make new released compiled against numpy v2

pgunn:
Right now the tensorflow pinning prevents us from getting a numpy 2.x in a normal environment build. We'll have to look at this again after the diffs land to get rid of tensorflow.

pgunn:
Now that we've removed tensorflow, and python has moved forward under us, it's time to give this much higher priority because newer versions of python don't often have pre-2.0 builds of numpy

pgunn:
Looks like those changed copy semantics are going to be something to work on.

https://github.com/scipy/scipy/pull/20172

pgunn:
And one other thing need adjusting too.

caiman.utils.utils.py:recursively_save_dict_contents_to_group() has a call to np.string_ which was removed. The docs suggest using np.bytes instead.

Otherwise we have several bits of the code all over the codebase that call the np.array() constructor with copy=False

(sidenote: the test_mrcnn codebase has some cuda related errors I need to look into, and the tests in there also should be modified to turn off verbose)

pgunn:
I think the above may fix the issues. The testing suggests I may need to do more work on the mrcnn tests.

Unclear if it's just the tests or the codebase, but the tests don't work well on my workstation which has a usable GPU. But that's unrelated to this issue.

pgunn:
Some of the C++/Python glue may need some adjusting too; I think this relates to #1550

pgunn:
Wondering if it's this:
https://numpy.org/devdocs/numpy_2_0_migration_guide.html#functionality-moved-to-headers-requiring-import-array

pgunn:
(some later fixes are in the provenance-fixing branch that came up in testing that)

pgunn:
[...]

PR #1549, opened 2025-12-16
Numpy 2.x

pgunn:
This tries to address the big issues with #1354, using the copy=None flag to the numpy.array constructor which is semantically close to the numpy 1.x behaviour (don't copy if you don't need to, but if you do don't error out). It also lifts the pin.

I don't think this works on numpy 1.x; we may actually need a minimum version pin, as it looks like the typedef for NPY_COPYMODE is a 2.x thing (the numpy codebase is not that easy to read though, both because these parts are a mix of C and python and because they use a lot of abstraction). The docs kind of imply that it's ok to pass none, but empirical tests suggest otherwise.

Numpy 1.x:
```
>>> import numpy as np
>>> data = np.array([[1,2,3,4],[2,3,4,5]])
>>> data
array([[1, 2, 3, 4],
       [2, 3, 4, 5]])
>>> data2 = np.array(data, copy=None)
  File "<stdin>", line 1, in <module>
ValueError: NoneType copy mode not allowed.
```

Numpy  2.x:
```
>>> import numpy as np
>>> data = np.array([[1,2,3,4],[2,3,4,5]])
>>> data2 = np.array(data, copy=None)
```

### DIFF
--- a/caiman/motion_correction.py
+++ b/caiman/motion_correction.py
@@ -1360,9 +1360,9 @@ def register_translation_3d(src_image, target_image, upsample_factor = 1,
     # real data needs to be fft'd.
     elif space.lower() == 'real':
         src_image_cpx = np.array(
-            src_image, dtype=np.complex64, copy=False)
+            src_image, dtype=np.complex64, copy=None)
         target_image_cpx = np.array(
-            target_image, dtype=np.complex64, copy=False)
+            target_image, dtype=np.complex64, copy=None)
         src_freq = np.fft.fftn(src_image_cpx)
         target_freq = np.fft.fftn(target_image_cpx)
     else:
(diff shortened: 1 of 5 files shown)


ethanbb:
Hey @pgunn, I didn't look at this before, but I was trying to test out the current code with numpy 1.x because I was seeing some changes in test results (for mesmerize) and wanted to see if numpy was the issue. But I ran into the fact that this change isn't compatible with numpy 1.x.

I don't know if you saw this part of the [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword), but they suggest using `np.asarray(...)` rather than `np.array(..., copy=None)` for this reason (the default behavior matches `np.array` with `copy=False` for 1.x and `copy=None` for 2.x). Do you think that would work? I'd like to make a PR to change these and remove that minimum pin, so we can compare the behavior between versions and also be compatible with packages that aren't on 2.x yet.

pgunn:
I'm open to np.asarray, but I'm wary of removing the minimum version pin because I think there were a few other things we had to change (C++ code, other things) and I think at least one of them was impossible to have both ways. Maybe I can look back through the changes and see if that's true or if I'm misremembering.

(I don't think removing the minimum version pin will actually make comparing the behaviour between versions easier - having two environments, one for an older caiman and one with the current, is already pretty easy, and I think also the unavailability of numpy 1.x builds for newer python versions is already forcing users off of numpy 1.x to some degree)

ethanbb:
> (I don't think removing the minimum version pin will actually make comparing the behaviour between versions easier - having two environments, one for an older caiman and one with the current, is already pretty easy, and I think also the unavailability of numpy 1.x builds for newer python versions is already forcing users off of numpy 1.x to some degree)

Ah yeah, I wasn't clear - I meant still using the latest caiman code, but with numpy 1.x in the environment instead. Although if what you say is true, that still won't work if I switch it to `asarray`. (Technically *I* can do that without changing the pin anyway since I have caiman installed through pip and conda doesn't complain if I install a package that doesn't match its requirements.)

PR #1551, opened 2025-12-18
CNMF/OnACID: First provenance system

pgunn:
For now this is meant as a conversation piece to decide what we think of this design of a provenance system; it hasn't seen any testing (although it builds), and the corresponding changes to OnACID have not been made


caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -132,6 +133,13 @@ def __init__(self, n_processes, k=5, gSig=[4, 4], gSiz=None, merge_thresh=0.8, p
         self.skip_refinement = skip_refinement
         self.remove_very_bad_comps = remove_very_bad_comps
 
+        self.provenance = [] # This will provide a rough record of the history of the object, largely with the intent of it
+                             # being useful in the serialized file form. The formatting for this will be a list of dicts,
+                             # the dicts all having at least the fields "event":str, "time":int, and "text":str
+                             # Other fields are permitted. Time is Unix epochtime. The semantics used here will be mirrored in
+                             # the OnACID class
+        self.provenance.append({'event': 'init', 'time': int(time.time()), 'text': 'CNMF Object initialized'})
```
ethanbb: 'text' seems too generic to me - how about 'description'? Or 'action', though that might be too similar to 'event'.

Also, should it be in a consistent tense? To match the others, this could be "Initialize CNMF object".

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -208,9 +216,13 @@ def fit_file(self, motion_correct=False, indices=None, include_eval=False):
         Returns:
             cnmf object with the current estimates
         """
+
+        # TODO: This both modifies self and returns a new cnmf object. Consider revising semantics.
```
ethanbb: I agree, since this does the initial fit and refit in one call, it doesn't make much sense to me for self to become the result after just the initial fit. Self should either remain unchanged, or become the re-fit object (in which case there should be no return value). To do the former, I think you could just deepcopy the object first. To do the latter, there would have to be some version of `refit` that worked in-place rather than returning a new CNMF object. But of course either way it would be a breaking change.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -290,6 +304,7 @@ def refit(self, images, dview=None):
             cnm
                 A new CNMF object
         """
+        self.provenance.append({'event': 'refit', 'time': int(time.time()), 'text': f'Run refit', 'data_target': str(images)})
         cnm = CNMF(self.params.patch['n_processes'], params=self.params, dview=dview)
```
ethanbb: It seems like the new object needs to inherit the provenance from self.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -316,6 +331,7 @@ def fit(self, images, indices=(slice(None), slice(None))) -> None:
         """
         logger = logging.getLogger("caiman")
         # Todo : to compartment
+        self.provenance.append({'event': 'fit', 'time': int(time.time()), 'text': f'Run fit', 'data_target': str(images)})
```
ethanbb: Is `str(images)` really useful? I think it trims to just show the beginning and the end of the movie, right? (Otherwise obviously this would be impractical.) 

I suppose comparing string representations could be used as a very rough heuristic for whether 2 movies were actually the same. To do so more rigorously would require iterating through the whole movie and computing a hash, I think, which is probably too costly for the benefit. So maybe this is a decent compromise, although it seems a bit hacky. Alternatively, maybe just save the first few frames as an array? Then at least you could compare those frames with another movie as floats, with whatever precision you wanted, rather than as strings.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -550,6 +566,7 @@ def remove_components(self, ind_rm):
                         indices of components to be removed
         """
 
+        self.provenance.append({'event': 'remove_components', 'time': int(time.time()), 'text': f'Remove named components', 'data_target': str(ind_rm)})
```
ethanbb: `data_target` seems too generic, and it's not really the same as the `data_target` in fit, so how about `inds_removed` or similar?

Also, how about 'Remove specified components' rather than 'named'? Maybe this is just me, but I think 'named' implies they have actual names (i.e., strings).

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -571,6 +588,8 @@ def compute_residuals(self, Yr) -> None:
              Yr :    np.ndarray
                      movie in format pixels (d) x frames (T)
         """
+        self.provenance.append({'event': 'compute_residuals', 'time': int(time.time()), 'text': f'Compute and store residuals inline'})
```
ethanbb: Can we be specific and say "Compute residuals and store in YrA"?

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -625,6 +644,8 @@ def deconvolve(self, p=None, method_deconvolution=None, bas_nonneg=None,
         args_in = [(F[jj], None, jj, None, None, None, None,
                     args) for jj in range(F.shape[0])]
 
+        self.provenance.append({'event': 'deconvolve', 'time': int(time.time()), 'text': f'Deconvolve on traces', 'method_used': str(method_deconvolution)})
```
ethanbb: How about 'method_deconvolution' instead of 'method_used' to match the parameter name?

ethanbb (changes requested):
Overall I think this field can provide some important information - it seems  like a good idea! I have some suggestions, mostly about naming & consistency.

pgunn:
> Overall I think this field can provide some important information - it seems like a good idea! I have some suggestions, mostly about naming & consistency.

Thanks - the main intent here is to provide some reasonably strong hints to the user how a given dataset came about; I think it may be impossible to chase it all the way to 100% reproducibility given sometimes people will reach into the objects through accessors, different versions of the algorithms, and even some amount of nondeterminism in the algorithms, but any improvement along these lines should be a plus if it doesn't cost too much in other parts of the design.

caiman/source_extraction/cnmf/cnmf.py
pgunn: I toyed with description as I wrote it; still open to it.  I don't have strong feelings between the two. Will revise in that direction.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Yeah, I did a change in this direction for fit() earlier; painful but worth it. It sucks that it has the potential to break notebooks as well as any code out there that builds on caiman, but I think the API being better-behaved is worth it.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Oh, whoops, good catch.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Oh, this is totally a goof on my part; wasn't reading carefully enough and thought images was a list of filenames as it is in some other methods. I wasn't meaning to include any kind of serialisation of data in this.

I haven't tested this code at all; probably would've been caught during testing.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Both sound fine. I was writing this with the idea that we'd try very hard to keep 'event' and any custom parameters stable over time, but I was going back and forth over where custom parameters should also be very custom to the event or try to bend towards being generic.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Yup.

caiman/source_extraction/cnmf/cnmf.py
pgunn: Sure

caiman/source_extraction/cnmf/cnmf.py
pgunn: Working through the changes; I imagine the normal use-case for this will be someone looking at an HDF5 file that's already been generated; past tense probably makes the most sense?

pgunn:
Addressed comments; need to make corresponding changes in online_cnmf and then start testing. Comments are welcome throughout the entire process.

pgunn:
One thing I haven't decided on; is it worth making library functions to handle making/reading provenance entries? There's some repeated code for writing them, and it might be nice to have some utility functions to read them, but the format isn't that hard to grok.

I added a file in the docs covering parts of the design.

pgunn:
Sorry that the changelog is getting so big; testing this has turned up a few more issues from the earlier numpy 2 switchover as well as some other things.

pgunn:
After (finally!) some testing, I need to modify caiman.utils.utils.recursively_save_dict_contents_to_group() to specifically handle provenance entries so they don't land as a dataset with fixed length that might clip them; I think if they land as an attribute they won't bump into that problem (will need to verify that). Another option is to use h5py.string_dtype, either way that's where I need to change the code.

pgunn:
Ok, remaining work is to modify the loaders (load_CNMF and load_OnlineCNMF) to properly handle these changes. Otherwise things look good.

pgunn:
@ethanbb @kushalkolar  This is getting closer to being ready to land; if either of you have the time to look it over for anything else you'd like to say, this would be a good time; all I need to do is handle making sure deserialization is handled correctly.

pgunn:
Verified that provenance survives serialization/deserialization

pgunn:
Oops, no, it does not. It lands as a string.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -290,7 +271,13 @@ def refit(self, images, dview=None):
             cnm
                 A new CNMF object
         """
-        cnm = CNMF(self.params.patch['n_processes'], params=self.params, dview=dview)
+        cnm = CNMF(self.params.patch['n_processes'], params=self.params, dview=dview) # New object; this call does NOT modify self
+
+        cnm.provenance += self.provenance # Hoping we don't need to check to see if it exists to cover deserialisation cases from old code
```
ethanbb: I think you're still working on this but just to make sure, we should verify this and update the comment

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -363,7 +363,7 @@ def initialize_components(Y, K=30, gSig=[5, 5], gSiz=None, ssub=1, tsub=1, nIter
             SC_sigma=SC_sigma, SC_use_NN=SC_use_NN, SC_nnn=SC_nnn,
             SC_normalize=SC_normalize, SC_thr=SC_thr)
 
-    elif method == 'pca_ica': # This code may be untested
+    elif method == 'pca_ica': # This code is undertested and may not work
```
ethanbb, line 366: should this be a logged warning rather than just a comment, to make it more discoverable?

caiman/source_extraction/cnmf/cnmf.py
pgunn: Realised that going through the constructor takes care of this. Removing the comment.

ethanbb:
Thanks for cleaning things up and adding documentation, even if it's not all related to this issue!

caiman/source_extraction/cnmf/initialization.py
pgunn: Makes sense; added a warning.

pgunn:
> Thanks for cleaning things up and adding documentation, even if it's not all related to this issue!

Important to keep fixing things we spot while doing the things we intend to do. I don't know if the codebase will ever be truly clean and pretty, but we can move it in the right direction.

pgunn:
Right now I am reasonably satisfied with the state of this PR; will likely merge in a few days unless I either find problems or get requests for changes soon.

PR #1555, opened 2026-01-12
Update scikit-image version to 0.25.0

Cilia-Jia:
# Description
Update scikit-image version to 0.25.0 due to the version 0.26 skimage.morphology.dilation() not having the 'shift_x' argument which in turn create a bug in colpy.spikepursuit.volspike.

Fixes # (issue)

Demo: 'demo_pipeline_voltage_imaging'

# Type of change

### DIFF
--- a/environment-minimal.yml
+++ b/environment-minimal.yml
@@ -22,7 +22,7 @@ dependencies:
 - pyside6
 - pytest
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image==0.25.0
 - scikit-learn >=1.2
 - scipy >= 1.10.1
 - tifffile
--- a/environment.yml
+++ b/environment.yml
@@ -31,7 +31,7 @@ dependencies:
 - pytest
 - pytest-cov
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image==0.25.0
 - scikit-learn >=1.2
 - scipy >= 1.10.1
 - tifffile
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "pytest",
   "pytest-cov",
   "torch",
-  "scikit-image >= 0.19.0",
+  "scikit-image==0.25.0",
   "scikit-learn >= 1.2",
   "scipy >= 1.10.1",
   "torchvision",


pgunn:
We could do this, but as a stopgap. We can't stay on an old pin forever; any thoughts on what the right longer-term fix would be?

(note that this needs corresponding changes to environment-minimal.yml, pyproject.toml, and the feedstock -- I can handle the last but for this PR it'd be good to catch the first two in the same go)

Cilia-Jia:
Hello, indeed it is only a stopgap. It is now also modified in environment-minimal.yml, pyproject.toml.

Here are my thoughts on a longer term solution:

### **The behaviour:**

From [scikit-image git](https://github.com/scikit-image/scikit-image/blob/v0.22.0/skimage/morphology/gray.py) Shift the binary image `footprint` in the left and/or up. This only affects 2D footprints with even number of rows or columns. It was no longer used in the v0.23 and totally removed from the args in v0.26

### **Our use**

The arguments **shift_x and shift_y** are only used in this line of **spikepursuit.py**:

> 193 # extract the context region from the entire movie
> 194 bwexp = dilation(bw, np.ones([args['context_size'], args['context_size']]), **shift_x=True, shift_y=True**)
> 195 Xinds = np.where(np.any(bwexp > 0, axis=1) > 0)[0]
> 196 Yinds = np.where(np.any(bwexp > 0, axis=0) > 0)[0]
> 197 bw = bw[Xinds[0]:Xinds[-1] + 1, Yinds[0]:Yinds[-1] + 1]

### **The fix**

If we want to keep the behaviour of shifting then we need to add it in the code but from what I see it seems unessecary. Indeed the line 197 crop the image, removing any shifting done. Therefore I would just remove the **shift_Y and shift_x**.

This should require testing that I had no time to do that's why i advice that we merge the quick win of block the version and then work on a more permanent solution with proper testing.

pgunn:
Thanks; much appreciated. I appreciate your looking into it. Will likely merge this soon (and make corresponding feedstock changes) and then look into the longer-term fix, unless I decide to go with your approach above instead (either way I'll make sure to give you thanks in the changelog for the next release).

pgunn:
@Cilia-Jia I did a separate PR for this that does similar logic to how the old shift_x/shift_y code worked, for that one call only. Feels like a middle ground between pinning old scipy and removing the code and relying on cropping.

You'll get credit either way in the changelog; does this look good to you? I've done a bit of testing, but if you'd like to give it some more exercise, the dev-volpy_no_shift_in_dilation has my change and you may be able to test it more throughly than I can (I don't have a lot of volpy-suitable data and this is an edge of the codebase I don't work with much).

If it looks good to you, I'll land it in dev and do a new release soon.

pgunn:
I'll probably merge my fix in a few days; let me know if you test it and run into issues.

environment-minimal.yml
```diff
@@ -22,7 +22,7 @@ dependencies:
 - pyside6
 - pytest
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image==0.25.0
```
bobqianic, line 25:
```suggestion
- scikit-image>=0.19.0,<0.26.0
```

environment.yml
```diff
@@ -31,7 +31,7 @@ dependencies:
 - pytest
 - pytest-cov
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image==0.25.0
```
bobqianic, line 34:
```suggestion
- scikit-image>=0.19.0,<0.26.0
```

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
   "pytest",
   "pytest-cov",
   "torch",
-  "scikit-image >= 0.19.0",
+  "scikit-image==0.25.0",
```
bobqianic, line 43:
```suggestion
  "scikit-image>=0.19.0,<0.26.0",
```

pgunn:
@bobqianic Hi, Thanks for looking at this, but I have an alternative approach that I prefer that reimplements the footprint-alteration that scikit-image used to do, rather than changing software pins; that's in a separate diff. Currently planning on merging that one instead, unless issues come up.

pgunn:
Fixed in dev in #1560

PR #1535, opened 2025-10-05
Caiman PyTorch -> Ver 1.0.1

manuelpaeza:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

# Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would break back-compatibility)
- [ ] Changes in documentation or new examples for demos and/or use cases.

# Branching
- All PRs should be made against the **dev** branch. The main branch is not often merged back to dev.
- If you want to get your PR out to the world faster (urgent bugfix), poke pgunn to cut a release; this will get it onto github and into conda faster

# Has your PR been tested?

If you're fixing a bug or introducing a new feature it is recommended you run the tests by typing

```caimanmanager test```

and

```caimanmanager demotest```

prior to submitting your pull request. 

Please describe any additional tests that you ran to verify your changes. If they are fast you can also
include them in the folder 'caiman/tests/` and name them `test_***.py` so they can be included in our lists of tests.

PR #1558, opened 2026-01-16
Test fixes related to pytorch & numpy

ethanbb:
# Description

This fixes the problem with test_mrcnn caused by test_keras on GPU torch installs that I brought up in a comment on #1535. It also fixes some remaining warnings/incorrect names from the numpy 2.0 transition that I noticed.

# Type of change

# Has your PR been tested?

Yes, on a Windows computer with the CUDA version of pytorch installed, all tests now pass. It would be ideal to have a test runner with GPU to use for CI, but it seems like that would incur extra charges if we use GitHub's (https://docs.github.com/en/actions/concepts/runners/larger-runners).

### DIFF
--- a/caiman/tests/test_keras.py
+++ b/caiman/tests/test_keras.py
@@ -12,9 +12,9 @@ try:
     import keras_core as keras
 except ImportError:
     import keras
-    
-def test_keras():
-    os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
+
+def test_keras(monkeypatch):
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", "-1")
 
     try:
         model_name = os.path.join(caiman_datadir(), 'model', 'cnn_model')
@@ -31,7 +31,4 @@ def test_keras():
     try:
         predictions = loaded_model.predict(A, batch_size=32)
     except:
-        raise Exception('NN model could not be deployed. use_keras = ' + str(use_keras))
-
-if __name__ == "__main__":
-    test_keras()
+        raise Exception('NN model could not be deployed.')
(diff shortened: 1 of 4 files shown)


caiman/tests/test_keras.py
```diff
@@ -12,9 +12,9 @@
     import keras_core as keras
 except ImportError:
     import keras
-    
-def test_keras():
-    os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
+
+def test_keras(monkeypatch):
```
pgunn, line 16: I'm not familiar with this syntax; what does this do?

caiman/utils/visualization.py
```diff
@@ -414,7 +414,7 @@ def get_contours(A, dims, thr=0.9, thr_method='nrg', swap_dim=False, slice_dim:
             if len(cumEn) == 0:
                 pars = dict(
                     coordinates=np.array([]),
-                    CoM=np.array([np.NaN, np.NaN]),
+                    CoM=np.array([np.nan, np.nan]),
```
pgunn, line 417: Interesting; I wasn't aware there was a preferred phrasing for this, but searching for it in the numpy docs I see they suggest people use nan.

pgunn: Looks like we also have np.NaN in caiman.utils.visualization - if you want to change it there too that'd be cool

caiman/tests/test_keras.py
ethanbb: I just learned it too, it's just a way of changing something for the duration of the test. See here: https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables

caiman/utils/visualization.py
ethanbb: Oh uh that is this file! And it's an error since numpy 2.0.

pgunn: Ah. Lol, definitely a sign that I should be getting to bed soon.

caiman/tests/test_keras.py
pgunn: Ah, very nice. Will definitely use this for other things.

pgunn:
verified that np.memmap works (np.core.memmap also works, but nowadays has a deprecation warning)

PR #1559, opened 2026-01-22
Fixing issues found with "ruff" linter

pgunn:
To try to catch more numpy 2.x issues, I'm giving the "ruff" linter a go (the numpy project recommended it); so far it's finding non-numpy things, but still good things to fix.


pgunn:
Working through what `ruff check --ignore=E722,F841,F401,F541,E741,E731,E402`

gives me. Maybe after this it'll be worth also running yapf and mypy.

PR #1560, opened 2026-01-22
For #1555, reimplement footprint logic that was shift_x/shift_y

pgunn:
Not very familiar with this code, but testing whether we can just remove the shift arguments (removed in newer scikit-learn versions) without needing to adjust the cropping logic. Intended to try to address #1555

Looking for inputs where this is True:
```
 (skimage.morphology.gray.dilation(data, shift_x=True, shift_y=True) != skimage.morphology.gray.dilation(data, shift_x=False, shift_y=False)).any()
```
(no dice so far)

### DIFF
--- a/caiman/source_extraction/volpy/spikepursuit.py
+++ b/caiman/source_extraction/volpy/spikepursuit.py
@@ -191,7 +191,7 @@ def volspike(pars):
         raise Exception('Dimensions of movie and ROIs do not accord')
         
     # extract the context region from the entire movie
-    bwexp = dilation(bw, np.ones([args['context_size'], args['context_size']]), shift_x=True, shift_y=True)
+    bwexp = dilation(bw, np.ones([args['context_size'], args['context_size']]))
     Xinds = np.where(np.any(bwexp > 0, axis=1) > 0)[0]
     Yinds = np.where(np.any(bwexp > 0, axis=0) > 0)[0]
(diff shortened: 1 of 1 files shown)


pgunn:
Ah, got it, one must provide a footprint to see a difference.

pgunn:
Emulate the old call, except as footprint preprocessing (so I don't need to essentially vendor retired code from scipy).

caiman/source_extraction/volpy/spikepursuit.py
```diff
@@ -191,7 +190,7 @@ def volspike(pars):
         raise Exception('Dimensions of movie and ROIs do not accord')
         
     # extract the context region from the entire movie
-    bwexp = dilation(bw, np.ones([args['context_size'], args['context_size']]), shift_x=True, shift_y=True)
+    bwexp = dilation(bw, _shift_footprint_old(np.ones([args['context_size']), args['context_size']]))
```
ethanbb, line 193: @pgunn this is a syntax error - is the end parenthesis supposed to be at the end of the line?

pgunn: Weird; I thought I had tested this code. I'll fix it in a few.

pgunn: Fixed in fa07d9855f4e6d2

pgunn: @ethanbb After this and after Ethan gets the volpy networks into a better state, I'll definitely need to do a release to get this out there.

PR #1564, opened 2026-02-02
Pixel_format procssing while load/save AVI/MKV movie 

zyyk78:
# Description
Added a function that can select pixel format(bgra or gray, related to 'isColor')  while saving movie 
Added a function that can convert pixel format to gray more accurately
Did not change its default behavior ( isColor = True)

# Type of change

# Branching
- All PRs should be made against the **dev** branch. The main branch is not often merged back to dev.

# Has your PR been tested?

```caimanmanager test```

and

```caimanmanager demotest```

### CODE REVIEW

File: caiman/base/timeseries.py [hunk 1]
```diff
@@ -247,9 +248,12 @@ def foo(i):
...
                 
             y, x = data[0].shape
-            vw = cv2.VideoWriter(file_name, codec, self.fr, (x, y), isColor=True)
+            vw = cv2.VideoWriter(file_name, codec, self.fr, (x, y), isColor=is_color)
             for d in data:
-                vw.write(cv2.cvtColor(d, cv2.COLOR_GRAY2BGR))
+                if is_color:
+                    vw.write(cv2.cvtColor(d, cv2.COLOR_GRAY2BGR))
+                else:
+                    vw.write(d)
```
pgunn, line 262: Are you sure this doesn't change behaviour if the caller doesn't override is_color?

caiman/base/movies.py
```diff
@@ -1151,7 +1151,17 @@ def load(file_name: Union[str, list[str]],
                     ret, frame = cap.read()
                     if not ret:
                         break
-                    input_arr[counter] = frame[:, :, 0]
+                    if frame.ndim == 2:
```
pgunn, line 1154: Sorry if this is a bit tedious, but could you add a comment block documenting the dimensions and shapes we're expecting and parsing out? Feels like this logic could use a bit of explanation; I can read it and get the gist, but it takes a bit.

caiman/base/timeseries.py
zyyk78: YES, is_color default  is true and it won't change its behavior (1-->3 channel , then saved as  FFV1+bgra (compress = 1 ) or RAW+yuv420p (compress = 0) ) 
while is_color is false, the cvt will not happen and will be saved as gray 
it will be saved as FFV1+gray / RAW+gray ( fourcc = GREY because iyuv doesn't support non-color, I dismissed this yesterday and now fixed)

caiman/base/movies.py
zyyk78: Sorry! I have added it and have removed some useless condition

PR #1565, opened 2026-02-03
Fix demos for windows (multiprocessing) + some misc. cleanup

ethanbb:
# Description

Various fixes for demotests, mainly so they run on Windows. This is long overdue (from me, I know I'm probably the heaviest Windows user who contributes).

The main issue with Windows is that all the demo scripts were missing the `if __name__ == '__main__':` guard. This is really important when using multiprocessing on Windows, because it uses the "spawn" method which relies on child processes being able to import the main file without running the actual script. Without the guard, one sees an endless sequence of these errors as the workers keep dying and the pool tries to replace them:

```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```
(since we're not freezing, just the `if __name__ == '__main__':` is sufficient.)

Other changes:
- I removed the `export MPLCONFIG=ps` line from the shell script and the equivalent for Windows. I cannot find any information about this variable online - it seems like it was probably supposed to be [`MPLBACKEND`](https://matplotlib.org/stable/users/explain/figure/backends.html#selecting-a-backend) instead. Nevertheless, the tests have been running fine without this setting the entire time they've existed - thanks to `xvfb-run`, an interactive backend works fine. I also confirmed that setting `MPLBACKEND=ps` instead of `MPLCONFIG` does not negate the need for `xvfb-run` (it's also needed for some opencv plotting).
- I removed the `block=True` from `matplotlib.pyplot.show(block=True)` and also ensured `plt.ion()` is always called, so that `show` doesn't block by default. This was causing certain runs to hang forever on Linux.
- To be consistent with the non-Windows run, I changed the Windows function to look for the demos under caiman_datadir rather than under the current directory.

# Type of change

# Has your PR been tested?

Yes!

### DIFF
--- a/caiman/caimanmanager.py
+++ b/caiman/caimanmanager.py
@@ -166,8 +166,7 @@ def do_nt_run_demotests(targdir: str) -> None:
     # Windows platform can't run shell scripts, and doing it in batch files
     # is a terrible idea. So we'll do a minimal implementation of run_demos for
     # windows inline here.
-    os.environ['MPLCONFIG'] = 'ps'             # Not sure this does anything on windows
-    demos = glob.glob('demos/general/*.py')    # Should still work on windows I think
+    demos = glob.glob('demos/general/*.py', root_dir=caiman_datadir())
     for demo in demos:
         print("========================================")
         print(f"Testing {demo}")
@@ -178,7 +177,8 @@ def do_nt_run_demotests(targdir: str) -> None:
         elif "demo_pipeline_voltage_imaging.py" in demo:
             print(f"  Skipping tests on {demo}: This needs Keras, an optional dependency")
         else:
-            out, err, ret = runcmd(["python", demo], ignore_error=False)
+            demo_path = os.path.join(caiman_datadir(), demo)
+            out, err, ret = runcmd(["python", demo_path], ignore_error=False)
             if ret != 0:
                 print(f"  Tests failed with returncode {ret}")
                 print(f"  Failed test is {demo}")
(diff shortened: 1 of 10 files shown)


eeshaanjum:
Tested this on macOS 14.7.1 (M1 Pro), Python 3.12.0, with the bundled demoMovie.tif.

On the 1.13.2 release (240e1f2e), this command never finished:

python demos/general/demo_pipeline.py --no_play --input ~/caiman_data/example_movies/demoMovie.tif

The parent process just sat there while worker processes kept getting respawned. I killed it after about two and a half hours.

On your branch (6a6fefc8), the same command finished in ~37s. I also tried adding just the if __name__ == '__main__': guard to the release script on its own, and that fixes it too (~21s), so the guard is what's actually doing the work.

That explains the macOS behavior as well: with the spawn start method, child processes re-import the main module, so without the guard the demo just starts running again in every worker. Looking at setup_cluster(), the macOS branch only switches to forkserver if it detects an IPython kernel. In a plain script there's no get_ipython(), the exception gets caught, and that workaround never kicks in.

Neither CI setup would've caught this either way: GitHub Actions only runs Linux, and the Jenkins macOS stage runs pytest but skips the demos, so this CLI path isn't covered on macOS anywhere. Thanks for the fix!

Issue #1110, opened 2023-06-13
Running the CNMF demo pipeline with p=2 causes the pipeline to hang on my system


loftusa:
Hi all,

I ran through the demo pipeline [here](https://github.com/flatironinstitute/CaImAn/blob/master/demos/notebooks/demo_pipeline.ipynb) with a few changed parameters.

I am running on a remote system with 32 cores and 100GB of memory and a GPU.
When I run 

```
# %%capture
#%% RUN CNMF ON PATCHES
# First extract spatial and temporal components on patches and combine them
# for this step deconvolution is turned off (p=0). If you want to have
# deconvolution within each patch change params.patch['p_patch'] to a
# nonzero value
cnm = cnmf.CNMF(n_processes, params=opts, dview=dview)
cnm = cnm.fit(images)
```

and deconvolution was turned on instead (e.g., set `p=2`), the run hung for about 3 hours before I killed it.
If this is intentional, there should likely be an error message preventing the pipeline to run at this point.

I tried messing around with `p_patch` as well, but that was with `p=2` still.

Here is my `opts`:

```
CNMFParams:

data:

{'caiman_version': '1.9.15',
 'decay_time': 1.8,
 'dims': (657, 657),
 'dxy': (1, 1),
 'fnames': ['/home/jovyan/work/notebooks/try_cnmf_exp/C11.tif'],
 'fr': 10,
 'last_commit': 'FILE-1686357314',
 'mmap_C': None,
 'mmap_F': None,
 'var_name_hdf5': 'mov'}

spatial_params:

{'block_size_spat': 5000,
 'dist': 3,
 'expandCore': array([[0, 0, 1, 0, 0],
       [0, 1, 1, 1, 0],
       [1, 1, 1, 1, 1],
       [0, 1, 1, 1, 0],
       [0, 0, 1, 0, 0]]),
 'extract_cc': True,
 'maxthr': 0.1,
 'medw': None,
 'method_exp': 'dilate',
 'method_ls': 'lasso_lars',
 'n_pixels_per_process': 13924,
 'nb': 1,
 'normalize_yyt_one': True,
 'nrgthr': 0.9999,
 'num_blocks_per_run_spat': 20,
 'se': None,
 'ss': None,
 'thr_method': 'nrg',
 'update_background_components': True}

temporal_params:

{'ITER': 2,
 'bas_nonneg': False,
 'block_size_temp': 5000,
 'fudge_factor': 0.96,
 'lags': 5,
[...]

PR #1566, opened 2026-02-04
Pydantic for CNMFParams

ethanbb:
# Description

(Note: this includes changes from #1565 as well.)

This adds automatic validation/conversion of the types of incoming parameters (whether from the constructor, JSON, or when loading a CNMF HDF5 file) and reversible serialization. With the power of pydantic, which I am myself only newly aware of, we can also do all kinds of semantic validation on the values as well (like ranges of ints/floats, making sure paths are valid, etc.); I didn't put much of that in here.

Benefits:
- Having the type declared for each parameter means that we can reliably convert data read from other formats (mainly JSON) to the right types, with the help of some custom validators and serializers. For example, the b'NoneType' and incorrect slice roundtripping can be totally handled by adding validation/serialization logic in one place (the params.py file).
- The type of each parameter is available for static type checking when accessing using dot syntax (now works because they are housed in dataclasses). I extended the base class for the sub-parameters to also work as a non-mutable mapping, so dict syntax will continue to work as well for backward compatibility (but not for setting, which was already not recommended in favor of the `change_params` method).
  - Each group params object can also be instantiated by itself with type-checked keyword arguments, and then passed as an argument to the `CNMFParams` constructor. This is a new alternative construction method that allows full type checking. I modified the `demo_pipeline.ipynb` notebook to demonstrate this method.
- I made it so that validation errors never actually raise an error, rather just print a warning that the type may be incorrect. I figured that this is more in keeping with the training-wheels-off nature of this package and the reality that tracking how each of them is used is difficult, thus setting the values to types I didn't anticipate may sometimes be necessary.
- Many of the checks previously done in `check_consistency` could be decentralized, reducing redundant checks when not relevant:
  - Most notably, defaults triggered by a `None` value and set based on other parameters were converted to properties on the respective `GroupParams` subclass. I set this up in a "clever" way so that any non-default value is stored in an actual field with an underscore, but with an "alias" of the public name, which allows it to be set by that name in the constructor. So if our field is `dependent`, the field `SomeParams._dependent` always contains `None` if the value has not been set and should take a default (or has been set explicitly to `None`). Meanwhile, if `some_params._dependent is None`, reading the param using `some_params.dependent` computes the correct default/fallback value in the property implementation. The benefit here is that if the params it depends on change, we know that the value of `dependent` is just a default and it will take an updated value, as opposed to the case where `_dependent is not None` indicating a user-set value.
  - For some params, such as `spatial.nb` and `temporal.nb`, there is no need for them to be set by the user; they are only ever computed/dependent (in this case, these should be read from `init.nb`). I converted these params to properties with no corresponding field, and attempting to set them again logs a warning (matching the existing "attempts to change them elsewhere are ignored" warning for `nb`).
  - Some checks can be handled by custom validators, such as converting to `None` from string or bytes types, or ensuring the values of `gSiz` are odd integers.
  - Some checks that only pertain to one parameter group could be moved to a model validator on that group rather than on the full `CNMFParams` object.

Overall I hope this leads to more robustness with the parameters system and make fixes to any residual bugs cleaner and less ad-hoc. I realize it is a new, somewhat complicated library to learn though - I'm happy to explain more and I'll leave some comments inline in the diff as well.

# Type of change

- [x] Changes in documentation or new examples for demos and/or use cases.

# Has your PR been tested?

Yes, `caimanmanager test` and `caimanmanager demotest` pass. I added a new test file for the new params features.


caiman/base/movies.py
```diff
@@ -1813,13 +1813,15 @@ def get_file_size(file_name, var_name_hdf5:str='mov') -> tuple[tuple, Union[int,
                 raise Exception('Unknown file type')
             dims = tuple(dims)
         else:
-            raise Exception('File not found!')
+            raise FileNotFoundError('File not found!')
```
ethanbb, line 1816: To catch more specifically and log a useful warning when checking data params

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -61,28 +62,9 @@ class CNMF(object):
...
-                 min_corr=.85, min_pnr=20, ring_size_factor=1.5,
-                 center_psf=False, use_dense=True, deconv_flag=True,
-                 simultaneously=False, n_refit=0, del_duplicates=False, N_samples_exceptionality=None,
-                 max_num_added=3, min_num_trial=2, thresh_CNN_noisy=0.5,
-                 fr=30, decay_time=0.4, min_SNR=2.5, ssub_B=2, init_iter=2,
-                 sniper_mode=False, use_peak_max=False, test_both=False,
-                 expected_comps=500, params=None):
+    """ 
+    def __init__(self, n_processes, dview=None, Ain=None, Cin=None, b_in=None, f_in=None,
+                 params: Optional[CNMFParams] = None, **param_kwargs):
```
ethanbb: I guess this cleanup does not depend on the new features, but passing params separately is deprecated anyway so I see no need to list them out here too.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -135,36 +109,26 @@ def __init__(self, n_processes, k=5, gSig=[4, 4], gSiz=None, merge_thresh=0.8, p
...
         self.estimates = Estimates(A=Ain, C=Cin, b=b_in, f=f_in,
                                    dims=self.params.data['dims'])
 
+    @property
+    def skip_refinement(self) -> bool:
+        return self.params.patch.skip_refinement        
+
+    @property
+    def remove_very_bad_comps(self) -> bool:
+        return self.params.patch.remove_very_bad_comps
```
ethanbb, line 143: Changed to properties so that they are recorded in the params object (I think for one of these, the value in the params object was previously never used, only the value passed here).

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -322,8 +285,9 @@ def fit(self, images, indices=(slice(None), slice(None))) -> None:
             logger.info("Parallel processing in a single patch is not available for data that is in memory or sliced")
 
         T = images.shape[0]
-        self.params.set('online', {'init_batch': T})
+        self.params.set('online', {'init_batch': T}, warn=False)
         self.dims = images.shape[1:]
+        self.estimates.dims = self.dims
```
ethanbb, line 311: This was not set before based on passed images (maybe is set in other places when needed) - caused a test failure and seemed inconsistent

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -395,14 +359,14 @@ def fit(self, images, indices=(slice(None), slice(None))) -> None:
                 return
 
             logger.info('update spatial ...')
-            self.update_spatial(Yr, use_init=True)
```
ethanbb, line 398: This `use_init` argument has not done anything since [2018](https://github.com/flatironinstitute/CaImAn/commit/5be88ecce793a348ed482076a788db4c70e875d4). Also, it was added to the spatial or temporal group of params even though it doesn't do anything (which isn't compatible with the new params behavior). I didn't remove the argument from `update_spatial` and `update_temporal`, but it now warns if set to something other than `None` and ignores it.

caiman/source_extraction/cnmf/cnmf.py
```diff
@@ -637,20 +601,21 @@ def deconvolve(self, p=None, method_deconvolution=None, bas_nonneg=None,
...
             Y:  np.array (d1*d2) x T
                 input data
-
         """
-        lc = locals()
-        pr = inspect.signature(self.update_temporal)
-        params = [k for k, v in pr.parameters.items() if '=' in str(v)]
-        kw2 = {k: lc[k] for k in params}
-        kwargs_new = {**kw2, **kwargs}
-        self.params.set('temporal', kwargs_new)
```
ethanbb, line 653: See above, all this messy inspection stuff is to add `use_init` to the temporal params, but `use_init` doesn't do anything anymore otherwise.

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -252,8 +254,8 @@ def initialize_components(Y, K=30, gSig=[5, 5], gSiz=None, ssub=1, tsub=1, nIter
         sn: ndarray
             per pixel noise
 
-        options_total: dict
-            the option dictionary
+        options_total: CNMFParams
+            full options object
```
ethanbb, line 258: I changed this for this method mainly to demo new attribute syntax/type checking for accessing param values

caiman/base/movies.py
```diff
@@ -1813,13 +1813,15 @@ def get_file_size(file_name, var_name_hdf5:str='mov') -> tuple[tuple, Union[int,
...
     elif isinstance(file_name, tuple):
         dims = caiman.base.movies.load(file_name[0], var_name_hdf5=var_name_hdf5).shape
         T = len(file_name)
 
     elif isinstance(file_name, list):
-        if len(file_name) == 1:
+        if len(file_name) == 0:
```
pgunn, line 1822: I keep going back and forth over whether the "more idiomatic" `if not len(...` or this is more clear; fine to see it this way or that way.

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -97,7 +99,7 @@ def downscale(Y, ds, opencv=False):
             ds = tuple(ds) + (1,)
 
         if d == 3 and Y.shape[-1] > 1 and ds[0] == ds[1]:
-            ds_mat = caiman.source_extraction.cnmf.utilities.decimation_matrix(Y.shape[:2], ds[0])
+            ds_mat = utilities.decimation_matrix(Y.shape[:2], ds[0])
```
pgunn: I'm not a fan of the shortened paths; this requires users to go read the import to see if this is a method call, a library called utilities, or a partial path and a call.

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -1154,9 +1155,9 @@ def HALS4shape(Yr, A, C, iters=2):
 
 
 @profile
-def greedyROI_corr(Y, Y_ds, max_number=None, gSiz=None, gSig=None, center_psf=True,
+def greedyROI_corr(Y, Y_ds, options: params.CNMFParams, max_number=None, gSiz=None, gSig=None, center_psf=True,
```
pgunn: Ordinarily I'd be worried about reordering params in case people pass in a large number of positionals, but this may be fine? We just have to hope no user code does that. It might be worth using the asterisk to stop use of positionals so it will throw a better error if any user code does that. We haven't been using asterisks in prototypes in the codebase but maybe it's time to start doing that?

caiman/source_extraction/cnmf/initialization.py
```diff
@@ -1479,14 +1481,14 @@ def init_neurons_corr_pnr(data, max_number=None, gSiz=15, gSig=None,
     # compute peak-to-noise ratio
     data_filtered -= data_filtered.mean(axis=0)
     data_max = np.max(data_filtered, axis=0)
-    noise_pixel = get_noise_fft(data_filtered.T, noise_method='mean')[0].T
+    noise_pixel = pre_processing.get_noise_fft(data_filtered.T, noise_method='mean')[0].T
```
pgunn: Apart from the paths being short, I like that there's some degree of pathing added.

caiman/source_extraction/cnmf/map_reduce.py
```diff
@@ -91,21 +92,22 @@ def cnmf_patches(args_in):
     slices.insert(0, slice(timesteps))
 
     images = np.reshape(Yr.T, [timesteps] + list(dims), order='F')
-    if params.get('patch', 'in_memory'):
+    if params.patch.in_memory:
```
pgunn, line 95: Like the nicer phrasing here.

caiman/source_extraction/cnmf/map_reduce.py
```diff
@@ -91,21 +92,22 @@ def cnmf_patches(args_in):
...
         images = np.array(images[tuple(slices)], dtype=np.float32)
     else:
-        images = images[slices]
+        images = images[tuple(slices)]
```
pgunn, line 98: What does this change cover?

caiman/source_extraction/cnmf/map_reduce.py
```diff
@@ -91,21 +92,22 @@ def cnmf_patches(args_in):
...
-        for group in ('init', 'temporal', 'spatial'):
-            opts.set(group, {'nb': params.get('patch', 'nb_patch')})
-        for group in ('preprocess', 'temporal'):
-            opts.set(group, {'p': params.get('patch', 'p_patch')})
+        opts.change_params({
+            'patch': {'n_processes': 1, 'rf': None, 'stride': None},
+            'init': {'nb': opts.patch.nb_patch},
+            'temporal': {'p': opts.patch.p_patch},
+            'preprocess': {'p': opts.patch.p_patch},
+        })
```
pgunn, line 109: The main thing going through this API will do is call check_consistency() which might do some followup changes on values; will need to think carefully about whether that will change behaviour in ways users might care about

caiman/source_extraction/cnmf/map_reduce.py
```diff
@@ -208,9 +210,10 @@ def run_CNMF_patches(file_name, shape, params, gnb=1, dview=None,
 
     params_copy = deepcopy(params)
     npx_per_proc = np.prod(rfs) // memory_fact
-    params_copy.set('preprocess', {'n_pixels_per_process': npx_per_proc})
-    params_copy.set('spatial', {'n_pixels_per_process': npx_per_proc})
-    params_copy.set('temporal', {'n_pixels_per_process': npx_per_proc})
```
pgunn, line 213: Was removing the temporal intentional?

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -1106,8 +1106,9 @@ def mc_next(self, t: int, frame: np.ndarray) -> np.ndarray:
                 # tile_and_correct[_3d] and motion_correct_iteration[_fast] return the shifts needed to apply to get the corrected frame
                 shift = tuple(-np.array(shift))
             else:
+                max_shift_ds = int(self.params.online.max_shifts_online / self.params.online.ds_factor)
                 frame_cor, shift = motion_correct_iteration_fast(
-                        frame, templ, *(self.params.get('online', 'max_shifts_online'),)*2)
+                        frame, templ, max_shift_ds, max_shift_ds)
```
pgunn, line 1111: What's this change doing?

caiman/source_extraction/cnmf/params.py
```diff
@@ -1,78 +1,779 @@
...
+@dataclass(kw_only=True, eq=False, frozen=True)
+class GroupParams(Mapping):
+    """
+    Struct that can also be used as a non-mutable mapping, to be used
+    for subfields of CNMFParams, which have historically been dicts.
+
+    Aliases and computed fields are used for parameters that are computed
+    from other fields (potentially elsewhere in CNMFParams) if an explicit
+    value is not provided. This allows these parameters to continue to be updated
+    when the params it depends on are changed. The field containing the user-provided
```
pgunn: Wow, this is super interesting.

test_demos.sh
```diff
@@ -32,8 +32,6 @@ if [ $OS == "Linux" ]; then
 	fi
 fi
 
-# Tell matplotlib to try to plot less to begin with by specifying a postscript backend
```
pgunn, line 35: What's this change?

demos/general/demo_pipeline_NWB.py
```diff
@@ -246,5 +246,6 @@ def handle_args():
     return parser.parse_args()
 
 ########
-main()
+if __name__ == '__main__':
```
pgunn, line 249: What's the reason for the __main__ stuff here? We're not expecting anyone to import the demos as a library, are we?

demos/general/demo_OnACID_mesoscope.py
```diff
@@ -135,4 +137,5 @@ def handle_args():
     return parser.parse_args()
 
 ########
-main()
\ No newline at end of file
+if __name__ == '__main__':
+    main()
```
pgunn: If we're touching this let's get that final newline in

demos/general/demo_OnACID_mesoscope.py
```diff
@@ -84,12 +84,16 @@ def main():
...
     T_shapes = 1e3*np.array(cnm.t_shapes)
     T_track  = 1e3*np.array(cnm.t_online) - T_motion - T_detect - T_shapes
-    plt.figure()
-    plt.stackplot(np.arange(len(T_motion)), T_motion, T_track, T_detect, T_shapes)
-    plt.legend(labels=['motion', 'tracking', 'detect', 'shapes'], loc=2)
-    plt.title('Processing time allocation')
-    plt.xlabel('Frame #')
-    plt.ylabel('Processing time [ms]')
+
+    if not cfg.no_play:
```
pgunn, line 88: Ah, good catch with this

caiman/motion_correction.py
```diff
@@ -1091,13 +1091,13 @@ def process_movie_parallel(arg_in):
 
             #            logger.debug('median computing')
             template = Yr.bin_median()
-            idx_dot = len(fname.split('.')[-1])
+            fname_no_ext = os.path.splitext(fname)[0]
             if save_hdf5:
-                Yr.save(fname[:-idx_dot] + 'hdf5')
-            np.savez(fname[:-idx_dot] + 'npz', shifts=shifts,
+                Yr.save(fname_no_ext + '.hdf5')
```
pgunn: Maybe an f-string?

caiman/tests/test_params.py
```diff
@@ -0,0 +1,422 @@
...
+from pydantic import ValidationError, TypeAdapter
+import pytest
+import scipy.special
+from tabulate import tabulate
+from typing import Any, cast
+
+from caiman.base import movies
+import caiman.utils.utils
+from caiman.paths import caiman_datadir
+from caiman.source_extraction.cnmf import params, cnmf
```
pgunn: alphabetise imports in this section

pgunn (commented):
In general I really like these changes (apart from the incomplete-path import names); this also strengthens the structure in the params object, something I've wanted to do for a long time.

For this I think it would need to be a mid-dot release because some user code might break with these changes, but we'll be glad it was done after the dust settles. Eventually someday it might make sense to start turning validation on, selectively.

It may be necessary to look over the volpy cloning of CNMFParams, as this will leave that codebase as the odd one out (and there's a chance this will break that); in the long run it probably makes sense to just pull the volpy parameters into the main params object and have volpy use that. Might not be necessary to do that in this PR (unless this breaks the volpy params handling).

caiman/source_extraction/cnmf/map_reduce.py
pgunn: Oh, actually I'm still working through the new reasoning in CNMFParams and my comment here may be incorrect

caiman/source_extraction/cnmf/online_cnmf.py
```diff
@@ -180,11 +180,11 @@ def _prepare_object(self, Yr, T, new_dims=None, idx_components=None):
         self.M = self.params.get('init', 'nb') + self.N
 
         if not self.params.get('online', 'update_num_comps'):
-            self.params.set('online', {'expected_comps': self.N})
```
ethanbb, line 183: I added a warning to the `set` method since it doesn't call `check_consistency`, but within caiman in most cases I just added "warn=False" to the `set` calls to avoid slowing things down/changing behavior unnecessarily. I don't think `set` is really necessary in the long run though (if anything it could be replaced by a "verbose" mode of `change_params`).

caiman/source_extraction/cnmf/params.py
```diff
@@ -1,78 +1,768 @@
...
+
+# deal with 'NoneType', b'NoneType' strings
+def interpret_string_none(obj: Any) -> Any:
+    if (isinstance(obj, str) and obj in ['None', 'NoneType']
+        or isinstance(obj, bytes) and obj in [b'None', b'NoneType']):
+        return None
+    return obj
+
+SafeNone = Annotated[None, BeforeValidator(interpret_string_none)]
+SafeAny = Annotated[Any, BeforeValidator(interpret_string_none)]
```
ethanbb, line 55: This is a common pattern in Pydantic. We want a way to signal some extra information about various fields in our object, for example custom ways of validating and serializing them. Typically this is done through the type annotation, although there's also a way to do it using a decorated method. However, we don't want to change the actual type of the field from the perspective of other Python functionality - for example, we wouldn't want to make it a subtype of None (if that's even possible?) or whatever the base type T is, because then type checkers wouldn't accept assignments of plain T's to the field. This is what `Annotated` is for.

Here I add the `interpret_string_none` function as a validator that runs before the main validation of the `None` and `Any` types, respectively. Then I use these annotated `SafeNone` and `SafeAny` types for fields where I want to add this string-None conversion step below (in this case, pretty much everywhere I would otherwise put None or Any), including within nested type structures.

caiman/source_extraction/cnmf/params.py
```diff
@@ -1,78 +1,768 @@
...
+    ['name'] mapping syntax, a property with the same name computes the actual value
+    to use if none has been provided.
+    """
+    __pydantic_config__ = ConfigDict(extra='forbid', serialize_by_alias=True)
+    __pydantic_fields__: ClassVar[Mapping[str, FieldInfo]]  # automatic, just declaring for typing purposes
+
+    group_name: ClassVar[str]  # name of the attribute on CNMFParams
+
+    # back-reference to help with some computed fields
+    _full_params: 'SkipJsonSchema[Optional[CNMFParams]]' = Field(default=None, init=False, exclude=True, repr=False)
```
ethanbb, line 132: I added this back-reference to the main object to help with parameters that are computed when left as default and whose defaults depend on parameters in other groups. Admittedly it is a bit hacky. We maintain an invariant that this is only non-None if the referred-to `CNMFParams` object has this object as one of its fields (i.e., `getattr(group_params._full_params, group_params.group_name) is group_params`). This means we have to make sure it's unset when copying, etc.

(Now thinking about it again, does this have to be true? As long as we're never modifying other params through this reference, maybe it's fine...it might be more surprising if computed properties are suddenly not available after making a copy.)

caiman/source_extraction/cnmf/params.py
```diff
@@ -1,78 +1,768 @@
...
+    when the params it depends on are changed. The field containing the user-provided
+    value (or None if none was provided) is prefixed with an underscore, but has
+    the un-prefixed name as an alias; this allows the name to be used in the constructor,
+    change_params, etc. This alias can also used to serialize the user-provided value
+    when the round_trip option is True (e.g., when saving to JSON, or when copying
+    the object using replace). When accessing the parameter using .<name> attribute or
+    ['name'] mapping syntax, a property with the same name computes the actual value
+    to use if none has been provided.
+    """
+    __pydantic_config__ = ConfigDict(extra='forbid', serialize_by_alias=True)
```
ethanbb, line 125: The 'forbid' mean unrecognized parameters will cause an error, but before we get to that point, the `check_for_extra_fields` model validator below finds any that don't match, removes them and logs a warning.

demos/general/demo_pipeline_NWB.py
ethanbb: See #1565 - it is important for multiprocessing when not using the fork method.

ethanbb (commented):
Submitting partial comments for now

caiman/source_extraction/cnmf/params.py
```diff
@@ -1,78 +1,779 @@
...
+import scipy.special
 from scipy.ndimage import generate_binary_structure, iterate_structure
-from typing import Optional
+from tabulate import tabulate
+from typing import (Optional, Any, Union, Literal, Annotated, Callable,
+                    Mapping, Iterator, TypeVar, ClassVar, cast, Type)
+import warnings
 
 import caiman.base.movies
 import caiman.utils.utils
```
pgunn: should be below source_extraction in import order - alphabetic

caiman/source_extraction/cnmf/initialization.py
ethanbb: OK, but pyright gives me a linting error the way that it was before with just `import caiman` (aside from the imports of specific functions): `"source_extraction" is not a known attribute of module "caiman"`. I think this kind of reference without explicitly importing the sub-package only works because the `__init__.py` of caiman, or any of the modules it imports, imports something within `caiman.source_extraction.cnmf.utilities`. So that's a fragile pattern even though it works most of the time.

I can change it to `import caiman.source_extraction.cnmf` and use absolute paths for the function calls again, though.

caiman/source_extraction/cnmf/initialization.py
ethanbb: Yeah, `params` isn't really an optional argument which is why I did this. Unfortunately that requires reordering the arguments. Alternatively an error could just be raised immediately if `params` is None.

Edit: I will do this

caiman/source_extraction/cnmf/map_reduce.py
ethanbb: Oh yeah, I guess this condition wasn't tested, but the index should definitely be a tuple of slices, like above, not a list. Indexing with a list of slices will give ``IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices`` (tested just now)

caiman/source_extraction/cnmf/map_reduce.py
ethanbb: No, you are correct I believe. I was a bit unsure when to switch things to `check_consistency` vs. keep `set`. In general though I think it's safer to check than not check, unless we are within the `check_consistency` method already?

caiman/source_extraction/cnmf/map_reduce.py
ethanbb: Yes, because `n_pixels_per_process` doesn't exist in the temporal group! Well, I guess it is added at this point currently, but searching shows it isn't referenced in `temporal.py` or `deconvolution.py`.

caiman/source_extraction/cnmf/online_cnmf.py
ethanbb: Just rewriting something a bit convoluted in a simpler way. The current version makes a 1-tuple out of the parameter value, doubles it into a 2-tuple, and then unpacks it into 2 parameters to the function.

test_demos.sh
ethanbb: See #1565

demos/general/demo_OnACID_mesoscope.py
ethanbb: This is also part of #1565 - I was hoping we could review/merge that first, so I didn't bother separating them

caiman/motion_correction.py
ethanbb: Sure, why not.

caiman/source_extraction/cnmf/params.py
ethanbb: Oh, don't we want to separate imports from the same package into their own section? I think that's how it's done in other files.

caiman/base/movies.py
ethanbb: I will say I prefer it this way! I didn't know the other way is considered more idiomatic.

caiman/source_extraction/cnmf/initialization.py
ethanbb: Or what about `from caiman.source_extraction import cnmf` and then use `cnmf.<whatever>`? I think this is the standard in Rust for the reason you stated.

Edit: I don't know what I was talking about, that wouldn't make sense. The Rust convention is to do what I did here, so it's `<module>.<function>()`

pgunn: I'm curious - do you get the linting error if you run the linter from outside the caiman source directory? e.g. if you have ~/src/caiman_dev

does

~$ mylinter src/caiman_dev/caiman/source_extraction/cnmf/CNMFParams.py

work? This might be one of those build quirks.

demos/general/demo_pipeline_NWB.py
pgunn: Ah, fair enough.

caiman/source_extraction/cnmf/params.py
pgunn: Oh, I highlit the wrong thing (or github isn't displaying this very well)

caiman/source_extraction/cnmf/initialization.py
ethanbb: Yeah, I think it is actually correct. I was surprised too. But I think the key to why it *normally* works is [here](https://docs.python.org/3/reference/import.html#submodules). Basically, any modules that have to be traversed in the initial package import (through `__init__.py`) are available through their fully-qualified names: "The invariant holding is that if you have sys.modules['spam'] and sys.modules['spam.foo'] (as you would after the above import), the latter must appear as the foo attribute of the former."

I tested and found that if I create a dummy package `foo` with an empty top-level `__init__.py` and a sub-package `bar`, doing:
```python
import foo
foo.bar
```
does cause a runtime error!

caiman/source_extraction/cnmf/params.py
ethanbb: Oh sorry missed that!

caiman/source_extraction/cnmf/initialization.py
pgunn: Hm. It would be nice to find a set of templates and practices so we never have to think about this again; if that means init.py imports everything, that seems fine. What do you think?

ethanbb:
That would be OK with me, if at every level the sub packages were all explicitly imported in the init file. Maybe this is out of scope for this PR though and I'll just revert it to how it was before here.

________________________________

pgunn:
> That would be OK with me, if at every level the sub packages were all explicitly imported in the init file. Maybe this is out of scope for this PR though and I'll just revert it to how it was before here.

I'll followup with a PR to do that after this lands.

Excited about this PR!

ethanbb:
(Oh wow, my direct email reply to that thread did not work correctly at all!)

>> That would be OK with me, if at every level the sub packages were all explicitly imported in the init file. Maybe this is out of scope for this PR though and I'll just revert it to how it was before here.
>
> I'll followup with a PR to do that after this lands.

Sounds good, however actually let's also check how it affects import time? Since that will recursively import every file, it might increase significantly, and `import caiman` is already pretty slow.

> Excited about this PR!

Me too! It was pretty tricky, I hope it's maintainable!

caiman/source_extraction/cnmf/params.py
ethanbb: Update - I removed the special-case copy and deepcopy methods because I realized they led to really surprising behavior when deep-copying the full `CNMFParams` object.

ethanbb:
I was also wondering about the parameter `p` which is in both the 'temporal' and 'preprocess' groups (as noted in #1153). It seems like, for the most part, the one in 'preprocess' takes precedence: in the main (non-patch) branch of `CNMF.fit`, we have this:

https://github.com/flatironinstitute/CaImAn/blob/487eef8859ad36e7040ed23ffc509f4a6d840a31/caiman/source_extraction/cnmf/cnmf.py#L400-L408

Later, in non-`skip_refinement` case, the value from `preprocess.p` is used in a second pass. So for this piece of code, `temporal.p` could be computed from `preprocess.p`, and there could be another argument to `update_temporal` to override `p` to 0 to handle the first pass.

The other times `update_temporal` is called, it's after patch processing and there's no explicit updating of `temporal.p` based on `preprocess.p`. This is a bit more troubling, because as far as I can tell `temporal.p` doesn't get updated through `run_CNMF_patches` either (the copies in each subprocess get updated to `p_patch`, but not the parent process params). So this could cause unexpected behavior if the two `p`s were different, but based on e.g. #1110 this is probably a bug and not intentional - and we should probably change it to read from `preprocess.p` as above.

Similar situation with the calls to `update_temporal_components` in preprocess.py - these happen before `temporal.p` is updated to match `preprocess.p`, but again I can't imagine that's intentional?

In the demos, as far as I can tell, they're always set together (and actually this would always happen automatically before switching to the nested params system). The only exception is the one for OnACID since that algorithm doesn't use `temporal.p` at all. So changing them to always be the same shouldn't affect the tests - maybe someone's weird custom pipeline settings would be affected but I think it would mostly reduce confusion.

Overall I think it would make sense to have `p` set on `temporal` and have `preprocess.p` be computed from it (and warn when trying to set it, like `nb`).

pgunn:
So, a bit of history here for context (I am not pushing for any particular solutions with this). Hoping I'm remembering this right.

A long time ago, there wasn't a params object, and the various CNMF and OnACID interfaces all accepted a lot of arguments, relying on convention and cludge to be coherent. We had a hack-a-thon way back when that implemented CNMFParams, hoping to turn the (relatively flat, disorganised) parameters into something more structured, leading to category systems for params, the Params object, and the long-time status quo. I think at the time we had hoped we'd have time to follow up and make it more consistent across the codebase, removing the long list of arguments, making OnACID and CNMF more consistent with each other, and possibly to pass CNMFParams deeper into the codebase rather than having it continue being a user-focused thing, but we never got around to doing that. One small advance I did a few years ago was to get more serious about the catgegories of params, gently nudging people to provide params as structured dicts rather than with a flat interface, and adding json interfaces. A former project member wanted to move things in a different direction, flattening all params, which burned some time and cooperation possibilities. So that's how we got here.

There are a lot of things we could do, further in the future, although I am very happy to see things like this diff that look like they'll improve code quality.

Future ideas:
* Integrate Volpy's old-style handling of params, merging those params into CNMFParams and having that codebase use CNMFParams 
* Have CNMF methods fully commit to CNMFParams in the way OnACID methods do
* Better handle and document recalculation of params (partly done by this diff)
* Possibly move some things that are in particular categories (often mirrored into others) into a global category
* Pass CNMFParams deeper than its current surface-level parts of the codebase, simplifying arguments (and argument passing) of some functions. This would make the global category easier to do because right now as CNMFParams often gets splatted out into the arguments to particular functions, not leaving the possibility of those functions looking at Params outside those targeting them (this is a tradeoff, as the current design enforces what code can look at what params, making analysis easier, at the cost of removing some flexibility we might like)
* Enforce typing for params (this diff is a step in the right direction for that)
* Choose between change_params() and set() and move everything over accordingly, leaving the older interface as deprecated for a while and then retired. This would take some work with code that integrates with caiman (from Pluvianus to mesmerize). Possibly add a "recalculate" (or similar) flag to whatever's kept to indicate whether to redo consistency fixups.
* Figure out if/how to record CNMFParams into the provenance system

ethanbb:
Hi Pat,

OK, that all makes sense, thanks for laying it out explicitly!

I realize that in this repository, there are a lot of things that could be designed better, and a full redesign would take a lot of time and effort that may not be worth it compared to adding new features or improving the algorithms. At the same time, there is a relatively small number of actual "bugs" where some combination of parameters or series of actions causes an unexpected or incorrect result. Not all of these bugs are currently caught by the tests, nor will they probably ever be, but in my view we should at least prioritize fixing the bugs that we can immediately recognize based on how the code is designed and that are most likely to be encountered, without regard for more sweeping design changes we might want to make eventually. (I'm sure you agree, just laying it out!)

In this PR, I tried to make a clean mechanism for merging what were historically treated as 2 or more separate parameters into one underlying changeable value, while maintaining backward compatibility with accessing them. Since that seems to be the situation with `p`, and there is the potential for it to cause subtle bugs if not set correctly currently, shouldn't we use this mechanism to fix it? (Also, while we're already committed to making this a potentially-breaking release).

pgunn:
> Hi Pat,
> 
> OK, that all makes sense, thanks for laying it out explicitly!
> 
> I realize that in this repository, there are a lot of things that could be designed better, and a full redesign would take a lot of time and effort that may not be worth it compared to adding new features or improving the algorithms. At the same time, there is a relatively small number of actual "bugs" where some combination of parameters or series of actions causes an unexpected or incorrect result. Not all of these bugs are currently caught by the tests, nor will they probably ever be, but in my view we should at least prioritize fixing the bugs that we can immediately recognize based on how the code is designed and that are most likely to be encountered, without regard for more sweeping design changes we might want to make eventually. (I'm sure you agree, just laying it out!)

I don't have a firm opinion on this; I've tried to fix bugs, but there's also some satisfaction in improving code quality, and those sweeping design changes have sometimes inspired me to spend more time on the project (my role has been very flexible and self-directed, so I've had this luxury). Fixing bugs is certainly a good thing to do though.

> In this PR, I tried to make a clean mechanism for merging what were historically treated as 2 or more separate parameters into one underlying changeable value, while maintaining backward compatibility with accessing them. Since that seems to be the situation with `p`, and there is the potential for it to cause subtle bugs if not set correctly currently, shouldn't we use this mechanism to fix it? (Also, while we're already committed to making this a potentially-breaking release).

I like the changes, even if my motivations are not exactly the same as yours. Eager to land this as soon as the current issues raised about volpy are sorted out and (after that) I have time to test/learn the code. I expect this will make things more maintainable.

ethanbb:
OK so in this latest commit, I made some more changes. I realized that instead of just rejecting changes to `nb` that were on the wrong group, it would be more likely to align with the user's intention to warn but still make the change on the correct group. On the other hand, if there are multiple *different* values for `nb` in one update, that should be an error rather than just taking the one on the `init` group. So I added some new logic to check for that.

Also, I set up `p` and `n_pixels_per_process` to work the same way, since they're both always set to be the same value anyway within `CNMF.fit`. I decided the canonical group for `p` should be `temporal` and for `n_pixels_per_process` should be spatial - even though they may be accessed more often through `preprocess`, the spatial and temporal groups seem to make more logical sense. Like `nb`, they can both still be read from the `preprocess` group.

Finally, I went through `CNMF.fit` and some of the functions it calls and tried to eliminate calls to `CNMFParams.set` that converted default placeholders to concrete values, in the same spirit as the parameters that are now computed dynamically instead of set to a default when not explicitly set. For example, `n_pixels_per_process` and `patch.stride` now remain None on the params object if not set to a non-None value. (`stride` could be one of those computed params, except that the `run_CNMF_patches` function uses a different way of computing the default, so I didn't want to commit to one way or the other in case it is ever run on its own.) 

This is a bit debatable - I've been thinking of the params object as something that might be reused in the future for other runs (e.g., for a grid search), so we should avoid changing parameters unexpectedly in ways that will change the behavior depending on the order of runs. But maybe this particular params object (which is saved with the CNMF results) should hold all the parameter values that were actually used, as a persistent record in case the code changes in the future? Now that I'm thinking about this some more, I think the right thing to do might be to just make a copy of any passed-in `CNMFParams` immediately and then treat that as something that can be freely modified to reflect the current run.

caiman/source_extraction/cnmf/params.py
```diff
@@ -660,368 +1341,207 @@ def __init__(self, fnames=None, dims=None, dxy=(1, 1),
...
+                               "normalize_init in group init automatically to False.")
+                self.set('init', {'normalize_init': False}, warn=False, verbose=False)
+            
+            # Set structuring element to the no-op value (previously done in initialization.greedyROI_corr)
+            ndim = len(self.data.dims) if self.data.dims is not None else 2
+            null_se = np.ones((1,) * ndim, dtype=np.uint8)
+            if not utilities.all_same(self.spatial.se, null_se):
+                logger.warning("using CNMF-E's ringmodel for background hence setting key "
+                               "se in group spatial automatically to null element.")
+                self.set('spatial', {'se': null_se}, warn=False, verbose=False)
```
ethanbb: I'm marking this as a draft for now because I found this behavior isn't quite consistent with the current behavior. Specifically, when initializing using `greedyROI_corr` (i.e., corr_pnr), there are 2 lines where `update_spatial_components` is called with the original value of `se` (lines 1245 and 1285) before it is set to this identity element for the call on line 1335 *and for the rest of the CNMF run*, since the params object gets modified. And I think this is causing a difference in outputs for one of the mesmerize tests.

I'd like to see if I can guess from the git history whether this is intentional or a mistake...since the current actual behavior is a bit weird, I don't really want to try to figure out a way to replicate it with an immutable params object if it might just be a bug.

(As an aside, I was thinking it might be good to incorporate the mesmerize tests or maybe something similar but not depending on mesmerize, as part of the caiman CI. Since they seems to be better at catching when the behavior of the algorithms changes, at least for the ones mesmerize uses.)

pgunn: I wish I knew whether that was intended offhand, but I don't.

Always up for more tests (provided they don't add dependencies)

ethanbb: I did some more reading and looking through the code... It looks like the first 2 times `update_spatial_components` is run during initialization, using the `se` provided, it is operating on data downsampled by `ssub` (by a factor of 2 by default). Then, in the 3rd `update_spatial` run during initialization and once more on the full image if a lower-rank background is requested (`nb` > 0, which is not recommended for 1p anyway), when operating on non-downsampled data, the closing step is skipped. I do think this is intentional, because `only_init` is always used for 1p, and the only other place where `update_spatial` would be called and implicitly use the 1x1 `se` set during `corr_pnr` initialization is when `only_init` is false.

 Since the effect of the closing operation depends on the scale of the image relative to the structuring element, it makes sense that the same `se` shouldn't be used for both downsampled and non-downampled data. To me, it would make more sense to use a downsampled `se` for downsampled data (e.g., 2x2 rather than 3x3) and the original `se` for original data. It's probably also fine to skip it on later spatial refinement runs, but it would be surprising if someone used a more aggressive downsampling and ended up filling in a large hole because a 3x3 closing step was still performed and then upsampled...so I think this is potentially worth changing to be consistent with the way the `se` is used for other init methods.

ethanbb: OK I think this is out of scope for the PR, so I documented it in #1575 and will restore the original behavior.

ethanbb:
Now passes mesmerize tests (with appropriate updates) 🎉

ethanbb:
Currently seems to be failing CI due to a certificate error for caiman.flatironinstitute.org.

PR #1571, opened 2026-02-19
Exclude scikit-image 0.26.0 due to remove_small_* regression

ethanbb:
# Description

Another scikit-image issue... In 0.26.0, the behavior of `remove_small_objects` and `remove_small_holes` was changed slightly, due to a parameter in each being deprecated and replaced with a different one (`min_size`) that effectively allows holes/objects that are 1 pixel larger. The change was documented, but imo still careless to change the behavior even when still using the deprecated parameter. Furthermore, it will be reversed in the next version (see this merged PR: scikit-image/scikit-image#8003).

It was causing a mysterious test failure on mesmerize-core. It seems reasonable to me to just exclude this specific version.

# Type of change

# Has your PR been tested?

Tested with mesmerize; I'm assuming caiman tests will still work with this version but CI will make sure

### DIFF
--- a/environment-minimal.yml
+++ b/environment-minimal.yml
@@ -22,7 +22,7 @@ dependencies:
 - pyside6
 - pytest
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image >=0.19.0, !=0.26.0
 - scikit-learn >=1.2
 - scipy >= 1.10.1
 - setuptools
--- a/environment.yml
+++ b/environment.yml
@@ -31,7 +31,7 @@ dependencies:
 - pytest
 - pytest-cov
 - pytorch
-- scikit-image >=0.19.0
+- scikit-image >=0.19.0, !=0.26.0
 - scikit-learn >=1.2
 - scipy >= 1.10.1
 - setuptools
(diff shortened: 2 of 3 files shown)


pgunn:
Could you also:
A) Do a PR to the feedstock
B) Change environment-minimal.yaml
C) Get pyproject.toml too

A would be a separate PR, but B and C can be part of this PR

ethanbb:
No problem, sorry I missed those.

pgunn:
Actually for the feedstock you can just commit directly (I think I gave you permissions?)

ethanbb:
So for the feedstock, how does that work in regard to releases? If I commit directly to the main branch, am I changing the current release without changing the version number? Is that allowed?

pgunn:
@ethanbb I think the build plumbing only is invoked when there's a new release in the caiman repo; I think that's what it polls.

ethanbb:
I'm still not exactly comfortable committing to the feedstock...I don't understand what that does. I'd prefer if you did it for this, or maybe if I have some extra time I'll try to read up on it

pgunn:
@ethanbb No problem.

PR #1572, opened 2026-02-19
Make compatible with numpy 1 again and fix some floating-point type promotion changes

ethanbb:
# Description

* Removes the minimum version pin on numpy 2.0 (discussed in #1549, we don't necessarily have to do this, but I confirmed it now works for me). All that was necessary to get it to run on numpy 1.26 was changing `np.array(..., copy=None)` to `np.asarray(...)`. 
* The remaining changes are to address changed behavior in numpy 2.0, all of which has to do with floating-point type promotion/conversion:
  *  New scalar type promotion behavior detailed here: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion. This is more predictable than the old behavior, but means that where sometimes previously a `float64` would be generated to avoid loss of precision, the new behavior keeps the output as a `float32`. I used the `np._set_promotion_state("weak_and_warn")` feature to track down cases that were causing differences in output, at least for CNMF(E) with params used in the mesmerize-core tests. So this is probably still not comprehensive; it will be hard to catch all of these.
  * Similarly, `np.fft` functions have new behavior with respect to floating-point types: https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-fft-support-for-different-precisions-and-in-place-calculations. Again, to maintain the old behavior, I ensured inputs to fft functions are `float64` or `complex128`.

I admit it's debatable whether we should always preserve old behavior rather than go with what numpy 2 now does by default. Probably a lot of the old behaviors maintained here were not intentional, and there are casts back to `float32` in various places to partially counteract the automatic conversion behavior. Nevertheless, the new behavior does sometimes cause a loss of precision, which could be important in some cases, and I think we should at least keep track of any changes in behavior so we know what's going on.

# Type of change

# Has your PR been tested?

Very much so!


caiman/source_extraction/cnmf/pre_processing.py
```diff
@@ -17,6 +17,7 @@
 import tempfile
 
 import caiman.mmapping
+from caiman.source_extraction.cnmf import utilities
```
pgunn: I'm not a big fan of this kind of import, but I can go back and adjust this more later

pgunn (commented):
I think this is broadly fine; I don't know how much effort should go into something that's already starting to fall off the end of the python build matrix, but if there are users out there that might benefit from this I don't object. Will merge this this week.

caiman/source_extraction/cnmf/pre_processing.py
ethanbb: Sorry I forgot you prefer more context. How about instead importing the `cnmf` sub-package and then using `cnmf.utilities.<function>`? Since we already have each CNMF module imported in that init file.

pgunn: That'd work; it's not about context so much as a preference to only think about either short function names or fully qualified ones; the middle ground weirds me out and keeps having me think there's a library called utilities.

pgunn:
Hey, sorry for the long delay. Things are settling down and I'm catching back up. Is this ready to merge?

ethanbb:
Hey, glad to hear it! This one should be, yes, I've been using it with these changes without any issues.

PR #1573, opened 2026-02-20
Clean up/consolidate make_contours/thresholding code and don't impose arbitrary order on tied pixels

### LINKED ISSUE #1570: nrg contour pixel sort order - dealing with ties
I'm still working through some numpy 1.x vs. 2.x differences to get mesmerize tests to pass. (Some of these are probably very trivial, but it's mostly floating point types not being promoted where they were previously, and there could be some places where the double precision is important, so I'm being cautious.)

One that I just realized this that the sort order for ties using e.g. `np.argsort` has changed (based on the notes, they switched to a faster implementation). We probably shouldn't rely on this specific order anyway, which is arbitrary. However, the cumulative-energy contour method currently cares about the order among ties, because tied pixels sorted as higher get a smaller cumulative energy number (accumulating from highest to lowest). We could just switch to the "stable" method, which would prevent issues in the future if the algorithm switches, but this strikes me as not the best solution because arbitrary ordering of tied pixels, stable or not, could lead to some pathological contours.

I think the most principled way is, as we accumulate energy from highest to lowest, for each set of tied pixels, add the total energy among all these pixels to each pixel (e.g., if the running total is 0.6, and the 3 highest-valued pixels not yet counted each have an energy of 0.05, each of them gets 0.6 + 3*0.05 = 0.75). My logic is that, for a given contour threshold (say 0.7), assuming some reasonable interpolation, we would add the same total pixel area compared to a threshold of 0.6 (2 pixels) but instead of picking 2 of them arbitrarily we would add 2/3 of each of the 3 that are tied.

I'm writing this to get the idea down first, then I will try implementing it and post some images here if it looks reasonable.

ethanbb:
# Description

I wrote a cumulative-energy normalization function (replacement for `norm_nrg` which was copied in a few places) that is more predictable than the existing one because it doesn't depend on the order numpy chooses to sort pixels using `np.argsort`. Instead, it uses `np.unique` and applies the total cumulative sum up to each pixel's squared value to each pixel with that value. (A little hard to describe - see the implementation).

One benefit is that because it's more predictable, it allows testing that the contours have stayed the same even if internal details of `np.argsort` change (as happened from numpy 1 to 2). This PR itself changes the contours, making it debatably a breaking change, but again I think the new version ("keeping ties") is less surprising. I have attached an example test script and figure showing the difference between the old and new methods. (Note, this case of a perfect Gaussian with many sets of perfectly tied pixels is an edge case - if there are no perfect ties, the contours will stay the same.)

contours_tie_example

[nrg_contour_test.py](https://github.com/user-attachments/files/25435822/nrg_contour_test.py)

Fixes #1570

# Type of change

- debatable: Breaking change (fix or feature that would break back-compatibility)

# Has your PR been tested?

`caimanmanager test` and `demotest` pass on linux with numpy 2.

### DIFF
--- a/caiman/base/rois.py
+++ b/caiman/base/rois.py
@@ -18,6 +18,7 @@ from skimage.segmentation import watershed
 import tempfile
 import time
 from typing import Any, Optional
+import warnings
 import zipfile
 
 from caiman.motion_correction import tile_and_correct, get_patch_centers, interpolate_shifts
@@ -57,7 +58,10 @@ def com(A, d1: int, d2: int, d3: Optional[int] = None, order: str = 'F') -> np.n
     coor = np.stack([c.ravel(order=order) for c in coors])
 
     # take weighted sum of pixel positions along each coordinate
-    cm = (coor @ A / A.sum(axis=0)).T
+    with warnings.catch_warnings():
+        # ignore 0 / 0 (should give nan)
+        warnings.filterwarnings('ignore', category=RuntimeWarning, message='invalid value encountered in divide')
+        cm = (coor @ A / A.sum(axis=0)).T
     return np.array(cm)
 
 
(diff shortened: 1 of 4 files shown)


pgunn:
Hi, wanted to check in on this; is this ready for merge?

ethanbb:
Yes this should be good to go @pgunn!

